PR #32, opened 2016-09-27
Vote bits sync

chappjc:
This implements vote bits (re)synchronization/repair.  Please test and review.  All of the new or modified functionality is tested on a testnet pool, but there may be more attack protections that should be made (walletRPCHandlerize any of this?).

In terms of behavior:
- On startup of dcrstakepool, a full vote bits sync of all tickets owned by the pool is performed.  It will be skipped if importscript gets run since that takes a long time.
- /tickets POST (set bits) has a short timeout so a large number of tickets does not block page loading.  TODO: implement a proper timeout with a channel and time.After select.
- A flash message will indicate that a tickets sync is in progress.  No tickets are shown in this period.
- /tickets GET checks for consistency of vote bits between wallets and launches a resync when needed.  A flash message is also used in this case, and no tickets are shown (as shown in image below).

Implementation:
- A RWMutex is used so multiple readers can access ticket vote bits data, while a single writer can lock before doing a set of any vote bits.
- An atomic is used to make a kind of "try lock" that the /tickets page can check without blocking.

It needs a dcrrpcclient version change for the new GetTickets RPC.

Issues:
- [x] `getstakeinfo`'s immature includes unmined, whereas `gettickets` does not include mempool. This prevents the startup check from passing [`int(gsi.Live+gsi.Immature) != numLiveTickets`](https://github.com/decred/dcrstakepool-private/pull/32/files#diff-454a23714e7d91a2a570cae4ce8ee807R1034) while tickets are in mempool. All is well after tickets get mined.  Workaround is to avoid such a check, and only check that the wallets agree with eachother w.r.t. gettickets.
- [x] On startup, if wallets are out of sync w.r.t. address indexes or scripts, then it is likely that `SyncVoteBits` will either report different numbers of tickets (before they are processing the import/sync) or report -4: syncing...  [here](https://github.com/decred/dcrstakepool-private/pull/32/files#diff-454a23714e7d91a2a570cae4ce8ee807R1311)  Rather than quit, the solution is to just skip the full ticket vote bits sync (check and repair).  The vote bits are repaired on-demand via /tickets.
- [x] Look like at least 33% of this PR is done.

Startup auto-repair:
resync

User changing vote bits, if process exceeds 3 seconds:
bitschange

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
-.vscode/
 config.toml
 dcrstakepool
 vendor
 .vscode/
-config.go*
+stock/
+votebitssync/
+controllers/config.go*
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ config.toml
 dcrstakepool
 vendor
 .vscode/
-stock/
-votebitssync/
 controllers/config.go*
+testing/
+*.orig
+
--- a/config.go
+++ b/config.go
@@ -32,7 +32,7 @@ const (
 	defaultDBPort           = 3306
 	defaultDBUser           = "stakepool"
 	defaultPoolEmail        = "admin@example.com"
-	defaultPoolFees         = 7.5
+	defaultPoolFees         = 5
 	defaultPoolLink         = "https://forum.decred.org/threads/rfp-6-setup-and-operate-10-stake-pools.1361/"
 	defaultRecaptchaSecret  = "6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe"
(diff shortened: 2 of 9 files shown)


jolan:
Can you please rebase this yet again? :kissing_heart:

We would like this functionality and it can finally be given the proper testing/attention now.

chappjc:
In the next hour or two, yup.

chappjc:
@jolan Oh, what do I do about my deleted repo?  Start a new PR from a new fork?

jolan:
Oh right :( Yeah that would work best.  I'll close this one.

PR #38, opened 2016-10-06
Vote bits control UX improvement

chappjc:
Addresses https://github.com/decred/dcrstakepool-private/issues/37, but this is not "the" solution, as pool control needs to be handled better internally.

Remove "pool control" checkbox since it implies state.  Instead, put another item in the drop down box and refer to it as "policy" instead of "control".  Setting pool control, especially with a checkbox, implies the pool is being _given_ the ability (control) to change vote bits as it sees fit.  Really it just applies the current pool policy when the button is clicked.  There are no future updates after this POST is handled.

**Before:**
before1
Say you uncheck it and click update.
before2
But then when it reloads the page, it's checked again.  Huh?  Not the right meaning.

**After:**

after
TODO: Have a "policy" set globally or in the controller rather than hard-coded in `TicketsPost`.  Then put this policy in in the drop down like "Use Pool Policy (Allow + custom bits)".  And a tooltip can inform what those custom bits are.

Note that this can be a pain to test on accounts with lots of tickets since setting vote its is still a bit broken (fixed in https://github.com/decred/dcrstakepool-private/pull/32).

### DIFF
--- a/views/tickets.html
+++ b/views/tickets.html
@@ -116,7 +116,7 @@
 			</tbody>
 			</table>
             <form method="post" class="form-inline" role="form">
-	            <div id="votebits" class="form-group" style="display: none;">
+	            <div id="votebits" class="form-group">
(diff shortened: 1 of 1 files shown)


chappjc:
Yeah, but not because of this.

Issue #28, opened 2016-08-02
accept4: too many open files -- when wallets inaccessible for extended period


chappjc:
On a few occasions when one or more wallets is inaccessible for an extended period, dcrstakepool will start giving the following error repeatedly:

```
http: Accept error: accept tcp 127.0.0.1:3000: accept4: too many open files; retrying in 1s
```

At this point you cannot quit with CTRL+C.  [Here is the SIGABRT trace](https://drive.google.com/open?id=0B0I-3DNmsdxFVmcxQll2d1k2MjQ).

alexlyp:
This problem should no longer occur with PR #55

Issue #11, opened 2016-04-19
Dynamic Wallet Discovery


themsquared:
This is more of an enhacement request.  Using something like consul.io, it would be great to be able to have optional "service discovery" for wallets. Rather than the recompile every time I need to add a wallet, I'd like to just register it with consul and have it dynamically be in use.

jolan:
Well we won't be implementing dynamic wallet discovery but would probably take a PR to do it.  Anyway, I'm closing this since it's not something we will ever work on.

PR #46, opened 2016-10-27
Add better wallet rpc connection handling

### LINKED ISSUE #17: status page looks good when stats doesn't respond
I just posted an issue with hung wallets here: https://github.com/decred/dcrwallet/issues/210

What I didn't say was that this machine is part of a stake pool.  One of those wallets was a wallet on the pool, so one of four wallets was not responding.

I forgot to mention in the other issue that they hung wallet returned some kind of client disconnected/lost error in response to getstake info, but returned fine with walletinfo.  However, walletinfo indicated daemonconnected false.

But I digress.   What I'm interested in here is having the pool's /status page return something other than RPC Status:  Running when in this situation.  Fortunately, I had a watch on /stats also which alerted me to this problem.  I had to go to the pool's machine and run a wallet check script, which runs both walletinfo and getstakeinfo on all 4 wallets.  Only getstakeinfo on the one wallet didn't work, and walletinfo for that one showed daemon not connected.

And this has been said before, but the other  pages really need to time out on the RPC calls and return something, anything.

alexlyp:
Fixes #17 and first pass for #28 and partly #11

### DIFF
--- a/config.go
+++ b/config.go
@@ -88,6 +88,7 @@ type config struct {
 	WalletPasswords  []string `long:"walletpasswords" description:"Pasword for wallet server"`
 	WalletCerts      []string `long:"walletcerts" description:"Certificate path for wallet server"`
 	Version          string
+	AdminIPs         []string `long:"adminips" description:"Expected admin host"`
 }
 
 // serviceOptions defines the configuration options for the daemon as a service on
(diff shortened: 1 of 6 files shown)


alexlyp:
Need to close and open a new PR since my private local repo borked and wasn't forked.

Moved to over here #55

PR #55, opened 2016-11-10
Add better wallet rpc connection handling

### LINKED ISSUE #17: status page looks good when stats doesn't respond

alexlyp:
Fixes #17 and first pass for #28 and partly #11

### DIFF
--- a/controllers/dcrclient.go
+++ b/controllers/dcrclient.go
@@ -185,7 +185,8 @@ type getStakeInfoMsg struct {
 
 // connectedResponse
 type connectedResponse struct {
-	err error
+	walletInfo []*dcrjson.WalletInfoResult
+	err        error
 }
 
 // connectedMsg
@@ -301,19 +302,38 @@ func (w *walletSvrManager) executeInSequence(fn functionName, msg interface{}) i
 	case getNewAddressFn:
 		resp := new(getNewAddressResponse)
 		addrs := make([]dcrutil.Address, w.serversLen, w.serversLen)
+		connectCount := 0
 		for i, s := range w.servers {
+			if w.servers[i] == nil {
+				continue
+			}
 			addr, err := s.GetNewAddress("default")
-			if err != nil {
+			if err != nil && (err != dcrrpcclient.ErrClientDisconnect &&
+				err != dcrrpcclient.ErrClientShutdown) {
 				log.Infof("getNewAddressFn failure on server %v: %v", i, err)
 				resp.err = err
 				return resp
+			} else if err != nil && (err == dcrrpcclient.ErrClientDisconnect ||
+				err == dcrrpcclient.ErrClientShutdown) {
+				addrs[i] = nil
+				continue
 			}
+			connectCount++
 			addrs[i] = addr
+		}
+
+		if connectCount < w.minServers {
+			log.Errorf("Unable to check any servers for getNewAddressFn")
+			resp.err = fmt.Errorf("not processing command; %v servers avail is below min of %v", connectCount, w.minServers)
+			return resp
 		}
 
 		for i := 0; i < w.serversLen; i++ {
 			if i == w.serversLen-1 {
 				break
+			}
+			if addrs[i] == nil || addrs[i+1] == nil {
+				continue
 			}
 			if !bytes.Equal(addrs[i].ScriptAddress(),
 				addrs[i+1].ScriptAddress()) {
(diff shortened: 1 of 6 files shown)


controllers/dcrclient.go
```diff
@@ -324,28 +344,52 @@ func (w *walletSvrManager) executeInSequence(fn functionName, msg interface{}) i
...
+				vawrs[i] = nil
+				continue
 			}
+			connectCount++
 			vawrs[i] = vawr
 		}
 
+		if connectCount < w.minServers {
+			log.Errorf("Unable to check any servers for validateAddressFn")
+			resp.err = fmt.Errorf("Unable to contact servers, please try again later")
```
jolan: How about something like:

```go
fmt.Errorf("not processing command; %v servers avail is below min of %v", connectCount, w.minServers)
```

controllers/main.go
```diff
@@ -1399,13 +1405,76 @@ func (controller *MainController) Stats(c web.C, r *http.Request) (string, int)
...
+
+	// Depending on how many wallets have been detected update RPCStatus.
+	// Admins can then use to monitor this page periodically and check status.
+	switch connectedWallets {
+	case 0:
+		rpcstatus = "Emergency!"
+	case 1:
+		rpcstatus = "Critical"
+	case 2:
+		rpcstatus = "Degraded"
```
jolan: Not really sure what to do about this now that minservers was added. Maybe something like:

```go
rpcstatus := "Unknown"
allWallets := len(walletInfo)

if connectedWallets == allWallets {
 rpcstatus = "OK"
} else {
  if connectedWallets == 0 {
   rpcstatus = "Emergency"
  } else if connectedWallets == 1 {
   rpcstatus = "Critical"
  } else {
   rpcstatus = "Degraded"
  }
}
```

sample-dcrstakepool.conf
```diff
@@ -55,3 +55,6 @@ testnet=1
 
 ; Useful for debugging RPC client issues.
 ;debuglevel=trace
+
+; Admin IPs to limit where certain pages can be accessed for admin maintainance (Status et al)
+;adminips=127.0.0.1
```
jolan: I would uncomment 127.0.0.1 since that's safe and it will bomb if it's not specified.  Also wrap at 80 lines. Maybe something like:

```
; Admin-specific functionality like the status page are limited by an IP address filter.
adminips=127.0.0.1
; Multiple values can be used and are separated by a comma.
;adminips=127.0.0.1,192.0.2.1,198.51.100.1
```

Also need to add minservers.

views/status.html
```diff
@@ -4,6 +4,44 @@
 		<h4 class="panel-title">Status</h4>
 	</div>
 	<div class="panel-body" style="background-color: #333; color: #ffffff;">
+	{{if not .WalletInfo}}
+	<p><span style="font-size: larger;"><b>Error receiving walletinfo</b></span></p>
+	{{end}}
+	{{if .WalletInfo}}
+	<table id="ticketslive" class="table table-condensed datatablesort">
```
jolan: Pasto.  table id should be waleltinfo or something

alexlyp:
@chappjc care to give this PR a shot and see if it avoids those issues you had mentioned on the other PR?

jolan:
Re-tested latest and works for me.  I think this should go in since it helps immensely and closes a lot of open issues.

Dyrk:
@jolan Why `r.RemoteAddr` is always "127.0.0.1" or "[::1]"?
Did you use some additional Nginx configuration to pass this RemoteAddr?
I see only this X-Real-IP header and it's probably not working with the RemoteAddr:
https://github.com/decred/dcrstakepool/blob/master/sample-nginx.conf#L54

jolan:
@Dyrk I am using the sample nginx.conf and the real IP is in RemoteAddr.  I'm using nginx 1.10.1.

chappjc:
I am seeing this issue too.  The verification email says "made from 127.0.0.1..."

PR #61, opened 2016-11-12
More sync and POST logging

chappjc:
More logging around startup sync, which can feel like it's hung for lots of tickets, and the POST handlers.

The calls to `GetTickets` are particularly slow, so it's helpful to have these messages:

```
08:11:00 2016-11-12 [INF] DCRS: Established connection to RPC server ...:...
08:11:00 2016-11-12 [INF] DCRS: Getting address index and importscript information from wallets.
08:11:01 2016-11-12 [INF] DCRS: Syncing wallets' redeem scripts and setting address indexes.
08:11:04 2016-11-12 [INF] DCRS: SyncVoteBits: Getting list of all live tickets on wallet 0.
08:12:09 2016-11-12 [INF] DCRS: SyncVoteBits: Excluding 0 unmined tickets in votebits sync.
08:12:09 2016-11-12 [INF] DCRS: SyncVoteBits: Checking number of tickets on wallet 1
...
08:14:29 2016-11-12 [INF] DCRS: Beginning resync of vote bits for [N] tickets.
08:14:30 2016-11-12 [INF] DCRS: Completed resync of vote bits for [N] tickets.
08:14:30 2016-11-12 [INF] DCRS: Starting wallet RPC manager
```

TODO: If a wallet votes during startup, thus decrementing the number of live tickets, this can be a frustration as it breaks the check and you are forced to try starting again.  The check for number of tickets [here](https://github.com/decred/dcrstakepool/blob/master/controllers/dcrclient.go#L1560) does not need to be done with `GetTickets`, although for wallet 0, it is necessary to use GetTickets to get the hashes that are passed to `SyncTicketsVoteBits` (a speedy function).  Use another way to check this or just don't bother with this check since there are other ticket syncs that happen before the vote bits sync.

Also disallow simultaneous password update and reset.

### DIFF
--- a/controllers/dcrclient.go
+++ b/controllers/dcrclient.go
@@ -1538,6 +1538,7 @@ func (w *walletSvrManager) SyncVoteBits() error {
 	}
 
 	// Check live tickets
+	log.Info("SyncVoteBits: Getting list of all live tickets on wallet 0.")
 	// legacyrpc.getTickets excludes spent tickets
 	ticketHashes, err := w.servers[0].GetTickets(true)
 	if err != nil {
@@ -1545,7 +1546,7 @@ func (w *walletSvrManager) SyncVoteBits() error {
 	}
 	ticketHashesMined := getMinedTickets(w.servers[0], ticketHashes)
 	numLiveTickets := len(ticketHashesMined)
-	log.Infof("Excluding %d unmined tickets in votebits sync.",
+	log.Infof("SyncVoteBits: Excluding %d unmined tickets in votebits sync.",
 		len(ticketHashes)-numLiveTickets)
 
 	// gsi, err := w.servers[0].GetStakeInfo()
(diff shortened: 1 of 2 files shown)


jolan:
Yep, I hit a case where 4 tickets got purchased right after the initial ticket count check happened so it blew up and I had to wait a minute or so and it restarted fine.

It is still a bit fragile but better that not syncing.  Could be moved to an admin page or be controlled with a config variable like syncAtStart.

Anyway, I'm logging improvements and enforcing only one form action works fine.

jolan:
Just hit it again:

14:27:30 2016-11-12 [INF] DCRS: Syncing wallets' redeem scripts and setting address indexes.
14:27:49 2016-11-12 [INF] DCRS: SyncVoteBits: Getting list of all live tickets on wallet 0.
14:28:21 2016-11-12 [INF] DCRS: SyncVoteBits: Excluding 0 unmined tickets in votebits sync.
14:28:21 2016-11-12 [INF] DCRS: SyncVoteBits: Checking number of tickets on wallet 1
14:28:35 2016-11-12 [ERR] DCRS: Non-equivalent number of tickets on servers 0, 1  (1514, 1511)
Fatal error in rpc sync: non equivalent num elements returned

chappjc:
Yeah, the number is live tickets in wallets >0 in that method should really be done without GetTickets.

PR #64, opened 2016-11-12
use headers then fallback to RemoteAddr when getting remote IP

### LINKED ISSUE #63: use X-Real-IP / X-Forwarded-For instead of remoteAddr
@Dyrk and @chappjc are reporting that remoteAddr always gives 127.0.0.1 due to nginx proxying.

It seems to work fine for me but it may be due to how to the environment I'm running it in.

Anyway, pretty simple to fix.

jolan:
Lightly tested and seems to work.

I also dos2unix'd the nginx.conf and changed X-Real-IP to X-Real-Ip since it was converting to that anyway due to the mime header specification.

I figured out why it worked for me -- I was running it on my firewall/nat box and it was spitting out my external address instead of 127.0.0.1 every time so it looked like it was working.  I tested with lynx from a VPS this time.

Fixes #63.

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -1,11 +1,13 @@
 package controllers
 
 import (
+	"bytes"
 	"crypto/rand"
 	"encoding/hex"
 	"errors"
 	"fmt"
 	"html/template"
+	"net"
 	"net/http"
 	"net/smtp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


jolan:
@dyrk @chappjc If you could review/test/confirm that'd be great.

controllers/main.go
```diff
@@ -28,6 +30,40 @@ import (
 	"gopkg.in/gorp.v1"
 )
 
+// ipRange is a structure that holds the start and end of a range of ip addresses
+type ipRange struct {
```
jrick: Wouldn't it be more appropriate/standard to use net.IPNet instead?  Then checking whether an IP is in a network is just `ipNet.Contains(ip)`.

chappjc:
This certainly might work, but for a reverse proxy through localhost, it seems like you can configure nginx (>=1.9) with the standard real_ip module like:

```
set_real_ip_from 127.0.0.1;
real_ip_header X-Forwarded-For;
real_ip_recursive on;
```

Then:

    proxy_set_header X-Real-IP $realip_remote_addr;

instead of

    proxy_set_header X-Real-IP $remote_addr;

nginx should work the magic for us.

jolan:
Yeah I'd prefer to just have nginx handle it but not totally sure I understand how to correctly set it up either.

chappjc:
Ah, my mistake was thinking http.Request.RealAddr comes from X-Real-IP.  This actually works:

https://github.com/chappjc/dcrstakepool/tree/realip

I can squash and PR.

chappjc:
Just another option:
https://github.com/decred/dcrstakepool/pull/66

jolan:
Closing in favor of #66.

PR #65, opened 2016-11-13
Speed up startup live ticket count check in SyncVoteBits

chappjc:
Since getMinedTickets was added, which removes unconfirmed tx from list from GetTickets, it's OK now to use getstakeinfo.(Live + Immature) to compare the total live (actually live+immature) ticket count on the other wallets with the count from getMinedTickets(gettickets(true)).

This makes startup much quicker and far less likely to hit a non-equiv error that you would get when live count goes down during startup.

Tested on testnet with and without unconfirmed tickets during startup (only time this is run).

- [x] test on mainnet
- [x] test once my testnet user gets live tickets too <s>(currently 0, all immature)</s> - looking good

### DIFF
--- a/config.go
+++ b/config.go
@@ -88,6 +88,7 @@ type config struct {
 	WalletUsers      []string `long:"walletusers" description:"Username for wallet server"`
 	WalletPasswords  []string `long:"walletpasswords" description:"Pasword for wallet server"`
 	WalletCerts      []string `long:"walletcerts" description:"Certificate path for wallet server"`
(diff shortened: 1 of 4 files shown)


chappjc:
Tested.

Still a log delay on the one GetTickets call (maybe put the fastest wallet machine first in the list), but no longer on the others.  Actual resync of vote bits on the listed tickets just takes a couple seconds for close to 2000 tickets.

chappjc:
Just added a switch, `--skipvotebitssync`, to skip the startup vote bits check and resync.  This does not prevent automatic resync on a per-user basis when desync is detected via /tickets page.

PR #66, opened 2016-11-14
Use ngx_http_realip_module and X-Real-IP.

chappjc:
An option for getting the "real" client IP into the app via the realip module of nginx.

Also update log_format with "$realip_remote_addr".

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"html/template"
+	"net"
 	"net/http"
 	"net/smtp"
 	"strconv"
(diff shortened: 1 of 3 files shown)


controllers/main.go
```diff
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"html/template"
+	"net"
```
jolan: This causes a compile error.

controllers/main.go
```diff
@@ -72,6 +73,31 @@ func randToken() string {
...
+			return parts[0]
+		}
+		return ip
+	}
+
+	realIP := r.Header.Get("X-Real-IP")
+	realIP = getHost(realIP)
+
+	if realIP == "" {
+		log.Infof(`"X-Real-IP" header invalid, using RemoteAddr instead (%s, %v).`,
```
jolan: Need to remove a format string or add another argument.

chappjc: No arguments needed, I think.

chappjc:
@jolan OK, done.  :)

PR #67, opened 2016-11-28
drop toml cfg; flag handling fix; logging cleanup

### LINKED ISSUE #30: Inconsistent logging backends
What logger should be used consistently?  In places it is btclog, glog, fmt.Printf...  Presumably dcrstakepoolLog (a btclog.Logger) is supposed to be used in most places, but is not.

### LINKED ISSUE #31: Command line switches are broken
`dcrstakepool -h` gives a nice summary of the options defined in config.go, but then if you try anything, you get a message from <s>Goji</s>`flag.Parse()` instead:

```
$ ./dcrstakepool -C ./dcrstakepool.conf
flag provided but not defined: -C
Usage of ./dcrstakepool:
  -alsologtostderr
        log to standard error as well as files
  -bind string
        Address to bind on. If this value has a colon, as in ":8000" or
        "127.0.0.1:9001", it will be treated as a TCP address. If it
        begins with a "/" or a ".", it will be treated as a path to a
        UNIX socket. If it begins with the string "fd@", as in "fd@3",
        it will be treated as a file descriptor (useful for use with
        systemd, for instance). If it begins with the string "einhorn@",
        as in "einhorn@0", the corresponding einhorn socket will be
        used. If an option is not explicitly passed, the implementation
        will automatically select among "einhorn@0" (Einhorn), "fd@3"
        (systemd), and ":8000" (fallback) based on its environment. (default ":8000")
  -config string
        Path to configuration file (default "config.toml")
  -log_backtrace_at value
        when logging hits line file:N, emit a stack trace
  -log_dir string
        If non-empty, write log files in this directory
  -logtostderr
        log to standard error instead of files
  -stderrthreshold value
        logs at or above this threshold go to stderr
  -v value
        log level for V logs
09:00:25 2016-09-26 [INF] DCRS: Version: 0.0.1-beta
  -vmodule value
        comma-separated list of pattern=N settings for file-filtered logging
```

On the upside, I use `-bind`, so it's good that I can do this.

btcsuite/go-flags should be the only command line parser.

Using glide.yaml and glide.lock in current dcrstakepool master.

### LINKED ISSUE #42: migrate config.toml options to dcrstakepool.conf
This is a remnant from the project we forked from and should be removed.

jolan:
This fixes all of the long standing bugs/inconsistencies with the stake pool code.

Closes #30.
Closes #31.
Closes #42.


server.go
```diff
@@ -69,20 +87,24 @@ func main() {
 
 	// Setup static files
 	static := web.New()
-	publicPath := application.Config.Get("general.public_path").(string)
 	static.Get("/assets/*", http.StripPrefix("/assets/",
-		http.FileServer(http.Dir(publicPath))))
```
chappjc: Could you check out my Nov 23 commit to the "stuff" branch?
https://github.com/chappjc/dcrstakepool/blob/0c54539aa59ff27cc17f0fcf05e92a45f1abc202/server.go#L60-L69

I managed to break the static files route when I moved off the default http router, and I realized that using `http.Handle` (two lines down) and using `server := &http.Server{Handler: app}` + `server.Serve()` at the end were incompatible.  Does this work for you?  I found it necessary to create a static files route `within` the `app` router.

```go
	// Setup static files
	publicPath := application.Config.Get("general.public_path").(string)
	assetHandler := http.StripPrefix("/assets/",
		http.FileServer(http.Dir(publicPath)))
	// static := web.New()
	// static.Get("/assets/*", assetHandler)

	// Apply middleware
	app := web.New()
	app.Handle("/assets/*", assetHandler)
```

glide.lock
```diff
@@ -70,8 +70,6 @@ imports:
   version: 8f6c67f0228a4a6a962008d6c3d1e694e4532e71
 - name: github.com/go-utils/uslice
```
chappjc: We can get rid of this too:
https://github.com/chappjc/dcrstakepool/commit/71a30638d7d5ded1856a0b0bfc9625c3e80204ba

server.go
```diff
@@ -36,15 +51,18 @@ func main() {
 	dcrstakepoolLog.Infof("Version: %s", version())
 	dcrstakepoolLog.Infof("Network: %s", activeNetParams.Params.Name)
 
-	filename := flag.String("config", "config.toml", "Path to configuration file")
+	// goji uses a bind flag in bind/bind.go, this allows go-flags to parse
+	// despite the conflict
+	flag.CommandLine.Init(os.Args[0], flag.ContinueOnError)
```
chappjc: See my comment from a few days ago on my branch:
https://github.com/chappjc/dcrstakepool/commit/0c2302bc0fcc94b0ba9dfce30ae3e53c68d63edb#commitcomment-19925636

chappjc: BTW, anything on that branch you'd want as a PR?
Perhaps [this](https://github.com/chappjc/dcrstakepool/commit/d8ee7e0b540e9519c4ae7daf2bb0e6f3323e133a) or the uslice removal?

jolan:
@chappjc Yeah if you want to PR your uslice removal and the signal handling/windows fix, that'd be great.  I'm not so sure about the stats one.  Ideally I'd like the stats page and stats retrieved from the API to be 100% consistent. (I don't think they are at the moment)

But please test this first if you can! I'd like to get this in soon, then once the API is improvements are finished and merged, I can pester all the pool operators to upgrade.

chappjc:
Consistency with the /stats page and the stats API response may be a contentious issue.  Most pools are changing up how they display the /stats page.

jolan:
I meant more the values for the various fields.  For example, my poolFees shows "7.5" on the webpage but "7.50" via stats. On second thought, I guess it doesn't matter much since hopefully everything will be using the API soon.

jolan:
Was hoping for some more tests but everyone appears rather busy at the moment.

I've tested it to the best of my ability on my development and "production" testnet stake pools and everything is working well.  In it goes.

chappjc:
Quick tests look good, but the log needs to be flushed before exiting or you don't get some errors.  Minor change: https://github.com/decred/dcrstakepool/pull/74

PR #68, opened 2016-11-30
Set Content-Type to json for API.

chappjc:
Not tested, but pretty straightforward since application.Route will apply any content-type in Env unless status is error.


chappjc:
superseded by https://github.com/decred/dcrstakepool/pull/73

PR #69, opened 2016-11-30
API with JSON types package

chappjc:
This is more for discussion than anything at this point as it is completely untested.  I believe the API needs several improvements, but this pertains to the following:

- For the convenience of encoding JSON by the server, struct response types with json tags should be created.
- For Go clients the json structs should put in their own package.
- The main API handler should be rewritten to use these types and the encodings/json package.
- The route handler should not include logic for parsing the version and command out of the url (goji and any router will do that).

Specifically, from the commit messages, this PR does the following:

- Add `poolapi` package with json structs, and new+modified api routes.  The `poolapi` package helps write go clients, and facilitate use of `json.Marshal` in generating the response for `http.ResponseWriter`.
- Modify `APIStats` and `APIPurchaseInfo` to use the new types and `json.Encoder.Encode()` directly into the `http.ResponseWriter`.
- `API()` -> `APIHandler` + `APIResponseHandler` + `APIInvalidHandler`.
- `APIHandler` uses `poolapi.NewResponse` to ready the response for `APIResponseHandler`.
- If no appropriate route+method exists, `APIInvalidHandler` is used.

A more elegant routing solution, like github.com/pressly/chi, should possibly be put in place, which could simplify how `APIHandler` is written.

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -177,7 +177,7 @@ func (controller *MainController) APIHandler(c web.C, w http.ResponseWriter,
 	switch r.Method {
 	case "GET":
 		switch command {
-		case "getpurchaseinfo":
+		case "getPurchaseInfo":
 			data, response, err = controller.APIPurchaseInfo(c, r)
 		case "startsession":
 			status, response = "success", "session started"
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -139,7 +139,8 @@ func NewMainController(params *chaincfg.Params, adminIPs []string, baseURL strin
 	return mc, nil
 }
 
-func (c *MainController) APIInvalidHandler(w http.ResponseWriter, r *http.Request) {
+// APIInvalidHandler responds to invalid requests. It satisfies http.Hander.
+func APIInvalidHandler(w http.ResponseWriter, _ *http.Request) {
 	resp := poolapi.Response{Status: "error",
 		Message: "invalid API version",
 	}
@@ -150,8 +151,9 @@ func (c *MainController) APIInvalidHandler(w http.ResponseWriter, r *http.Reques
 	}
 }
 
-func APIResponseHandler(resp *poolapi.Response, code int, w http.ResponseWriter,
-	r *http.Request) {
+// WriteAPIResponse marshals the given poolapi.Response into the
+// http.ResponseWriter and sets HTTP status code.
+func WriteAPIResponse(resp *poolapi.Response, code int, w http.ResponseWriter) {
 	w.Header().Set("Content-Type", "application/json; charset=utf-8")
 	w.WriteHeader(code)
 	if err := json.NewEncoder(w).Encode(resp); err != nil {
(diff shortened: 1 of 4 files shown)


jrick:
Totally on board with this concept (but haven't looked at the code yet).  Nearly had a heart attack when I looked at how the HTTP api was implemented (since there was no documentation, I needed to look at the implementation to see what kinds of objects I was dealing with when adding the [client api support to Paymetheus](https://github.com/decred/Paymetheus/tree/master/Paymetheus.StakePoolIntegration)).

chappjc:
This was the least huge change while keeping most of the scaffolding and without changing the returned objects.  I'd ideally make more sweeping changes.

poolapi/apijson.go
```diff
@@ -0,0 +1,41 @@
+package poolapi
+
+type Response struct {
+	Status  string      `json:"status"`
+	Message string      `json:"message"`
+	Data    interface{} `json:"data,omitempty"`
```
jrick: I think the data is always a json object with json strings as values.  In go that would be a map[string]string.

poolapi/apijson.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func NewResponse(status, message string, data interface{}) *Response {
+	return &Response{status, message, data}
+}
+
+type PurchaseInfo struct {
+	PoolAddress   string `json:"pool_address"`
```
jrick: this is a breaking change. the previous api did not have underscores in these keys.

poolapi/apijson.go
chappjc: Ah, yes.

controllers/main.go
```diff
@@ -137,55 +139,67 @@ func NewMainController(params *chaincfg.Params, adminIPs []string, baseURL strin
...
+	command := c.URLParams["command"]
+
+	httpStatus := http.StatusOK
+	var aresp *poolapi.Response
 
 	switch r.Method {
 	case "GET":
 		switch command {
-		case "getPurchaseInfo":
+		case "getpurchaseinfo":
```
jrick: this is another breaking change

controllers/main.go
```diff
@@ -137,55 +139,67 @@ func NewMainController(params *chaincfg.Params, adminIPs []string, baseURL strin
...
 			data, response, err := controller.APIPurchaseInfo(c, r)
-			return APIResponse(data, response, err), http.StatusOK
+			aresp = poolapi.NewResponse(err.Error(), response, data)
```
jrick: panics if err is nil.  the status must be "success" when there is no error

poolapi/apijson.go
chappjc: Right, I did this in several spots as it seemed more idiomatic json, but I'm not strongly in favor of this.

controllers/main.go
chappjc: Yeah, clearly untested.  :) Gonna try and refine today

controllers/main.go
jrick: and fwiw, i'm ok with making the api more consistent by doing this, but please do it in a different commit with a bump to the version.  refactoring prs shouldn't change observable behavior.

chappjc:
Documentation needed all over if we go this way

chappjc:
@jrick Thanks for the feedback!  jolan mentioned the imminent needed for the API, so I wanted to get a detailed discussion going before it becomes hard to make breaking changes.

jolan:
I just hacked up something quickly since I was told we needed it ASAP for 0.6.0 and that ended up not being the case so I'm definitely onboard with these improvements.

chappjc:
@jolan And it did the trick!  Getting started is always the hardest part.  Not judging.  ;)

poolapi/apijson.go
chappjc: Actually, I'm not sure how this would work though.  See `NewResponse`.  The two types presently passed to `data` are the two other new structs `PurchaseInfo` and `Stats`.  (Not maps).

chappjc: Although I get how a map[string]interface{} is useful for unmarshalling arbitrary json data.

chappjc:
@jrick, @jolan Reverted breaking changes, and fixed APIHandler's issues, hopefully.

poolapi/apijson.go
jrick: and both of those are json objects with json string values.  When the json for either response is unmarshaled into this Response type it will already be unmarshaled into a map[string]string package into the empty interface.  Typing it correctly just removes the type assertion and mandates that any responses we add later follow the same format.

You could also make it a *json.RawMessage which will keep the data in a byte slice so it can be unmarshaled to the expected type.  There's no one-size-fits-all solution here and I could see clients wanting both types.

jrick: also, yeah I would choose map[string]interface{} over map[string]string here. don't want to limit ourselves to only string values.  future versions of the api should use more appropriate json types.

jrick: if a client unmashals a response, and this struct has the field `Data interface{}` the underlying concrete type will be a `map[string]interface{}`.  The json decoder has no idea what the expected data is so it doesn't know it should be unmarshaled into a `PurchaseInfo` or `Stats`.

if the field is a `*json.RawMessage` then it remains a byte slice and the client is able to do a second unmarshal to fill a `Stats` or `PurchaseInfo` or whatever else we end up adding.  This is both faster and easier than going from json -> `map[string]interface{}` -> `ExpectedType`.

Making it a `interface{}` is the easiest type to use on the server side (just set the field to the `Stats` or `PurchaseInfo` and let the json package do the rest), and does not enforce that all results are json objects, which may or may not be useful depending on what other methods are added later.

So `interface{}` is probably best to use here, on the server end.  Clients will probably want it to be `*json.RawMessage`.  Clients can either declare that type themselves or we can provide it for them along with the server's type.

The server's type can probably also be unexported, since clients won't find a whole lot of use from it.

chappjc: I'm not gonna write any client functionality in the PR, but my understanding is it needs to work sort of like this: https://www.socketloop.com/tutorials/golang-marshal-and-unmarshal-json-rawmessage-struct-example.  While we probably don't need an explicit `"type"` key, it might be nice to have more that just the status.

jrick: Sure. interface{} would be the best option here on the server's type.  I still think it should be unexported as we don't expect clients to use it.

chappjc:
Well, this was good discussion, but things need to be reorganized a bunch.

Functions that actually write to http.ResponseWriter should not be in the `controller` package, especially not in a controller method.  The `system` package is responsible for the actual network responses.  So, the `Application` type (in `system`) should probably get an "`APIHandler`" method that gets the response _components_ from the current `controller.APIHander`, creates the actual json response struct, and calls the write response helpers.  These helpers are `WriteAPIResponse` and `APIInvalidHandler`, which should also be in "system".
The server-use-only response struct can live in `system` too, separate from `apijson` but still using `poolapi.Stats` and `poolapi.PurchaseInfo` for the `Data` field.

PR #71, opened 2016-11-30
Portable ReloadTemplatesSig

chappjc:
Let it build on Windows, but this disables template reloading on CTRL+C.


chappjc:
@jolan rebased, added the approach used in bravo branch deal with the new templates path option (make it a field of Application)

jolan:
Thank you again :+1:

PR #73, opened 2016-12-01
	Add poolapi package with json structs, new api routes.

chappjc:
Trying PR https://github.com/decred/dcrstakepool/pull/69 again.

**UPDATED: Pretty well tested**, but needs more documentation.

I believe the API needs several improvements, but this pertains to the following:

- For the convenience of encoding JSON by the server, struct response types with json tags should be created.
- For Go clients the json structs should put in their own package.
- The main API handler should be rewritten to use these types and the encodings/json package.
- The route handler should not include logic for parsing the version and command out of the url (goji and any router will do that).

Specifically, from the commit messages, this PR does the following:

- Add `poolapi` package with json structs, and new+modified api routes.  The `poolapi` package helps write go clients, and facilitate use of `json.Marshal` in generating the response for `http.ResponseWriter`.
- Modify `APIStats` and `APIPurchaseInfo` to use the new types and `json.Encoder.Encode()` directly into the `http.ResponseWriter`.
See the commit for more details.

A more elegant routing solution, like github.com/pressly/chi, should possibly be put in place, which could simplify how `APIHandler` is written.

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -22,6 +22,7 @@ import (
 	"github.com/decred/dcrd/blockchain/stake"
 	"github.com/decred/dcrstakepool/helpers"
 	"github.com/decred/dcrstakepool/models"
+	"github.com/decred/dcrstakepool/poolapi"
 	"github.com/decred/dcrstakepool/system"
 	"github.com/haisum/recaptcha"
 	"github.com/zenazn/goji/web"
@@ -138,54 +139,45 @@ func NewMainController(params *chaincfg.Params, adminIPs []string, baseURL strin
 }
 
 // API is the main frontend that handles all API requests.
-// XXX This is very hacky.  It can be made more elegant by improving or
-// re-writing the middleware and switching to a more modern framework
-// such as goji2 or something else that utilizes the context package/built-in.
-// This would make it much easier to share code between the web UI and JSON
-// API with less duplication.
-func (controller *MainController) API(c web.C, r *http.Request) (string, int) {
-	// we expect /api/vX.YZ/command
-	URIparts := strings.Split(r.RequestURI, "/")
-	if len(URIparts) != 4 {
-		return "{\"status\": \"error\"," +
-			"\"message\": \"invalid API request\"}\n", http.StatusOK
-	}
-	version := URIparts[2]
-	command := URIparts[3]
+func (controller *MainController) API(c web.C, r *http.Request) *system.APIResponse {
+	command := c.URLParams["command"]
 
-	if version != "v0.1" {
-		return "{\"status\": \"error\"," +
-			"\"message\": \"invalid API version\"}\n", http.StatusOK
-	}
+	// poolapi.Response comprises a status, message, and a data struct
+	var response, status string
+	var data interface{}
+
+	var err error
 
 	switch r.Method {
 	case "GET":
 		switch command {
 		case "getPurchaseInfo":
-			data, response, err := controller.APIPurchaseInfo(c, r)
-			return APIResponse(data, response, err), http.StatusOK
+			data, response, err = controller.APIPurchaseInfo(c, r)
 		case "startsession":
-			return APIResponse(nil, "session started", nil), http.StatusOK
+			status, response = "success", "session started"
 		case "stats":
-			data, response, err := controller.APIStats(c, r)
-			return APIResponse(data, response, err), http.StatusOK
+			data, response, err = controller.APIStats(c, r)
+		default:
+			return nil
 		}
 	case "POST":
 		switch command {
 		case "address":
-			_, response, err := controller.APIAddress(c, r)
-			return APIResponse(nil, response, err), http.StatusOK
+			_, response, err = controller.APIAddress(c, r)
 		case "signin":
-			_, response, err := controller.APISignIn(c, r)
-			return APIResponse(nil, response, err), http.StatusOK
+			_, response, err = controller.APISignIn(c, r)
 		case "signup":
-			_, response, err := controller.APISignUp(c, r)
-			return APIResponse(nil, response, err), http.StatusOK
+			_, response, err = controller.APISignUp(c, r)
+		default:
+			return nil
 		}
 	}
 
-	return "{\"status\": \"error\"," +
-		"\"message\": \"invalid API command\"}\n", http.StatusOK
+	if err != nil {
+		status = err.Error()
+	}
+
+	return system.NewAPIResponse(status, response, data)
 }
 
 // APIAddress is AddressPost API'd a bit
@@ -327,11 +319,10 @@ func APIResponse(data []string, response string, err error) string {
 }
 
 // APIPurchaseInfo fetches and returns the user's info or an error
-func (controller *MainController) APIPurchaseInfo(c web.C, r *http.Request) ([]string, string, error) {
+func (controller *MainController) APIPurchaseInfo(c web.C,
+	r *http.Request) (*poolapi.PurchaseInfo, string, error) {
 	session := controller.GetSession(c)
 	dbMap := controller.GetDbMap(c)
-
-	var purchaseInfo []string
 
 	if session.Values["UserId"] == nil {
 		return nil, "purchaseinfo error", errors.New("invalid session")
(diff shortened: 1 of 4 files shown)


poolapi/apijson.go
```diff
@@ -0,0 +1,43 @@
+package poolapi
+
+import (
+	"encoding/json"
+)
+
+type Response struct {
```
chappjc: Maybe this needs a "type" field so the client can double check the `Data` field is what they expect based on the request?

jrick: Client should be capable of mapping the response directly back to the request, so it knows what the expected data would be.

system/core.go
```diff
@@ -91,7 +100,7 @@ func (application *Application) Close() {
 	glog.Info("Bye!")
 }
 
-func (application *Application) Route(controller interface{}, route string) interface{} {
```
chappjc: Despite this one improvement (as I see it) here, I'm not fond of how `Route` works, particularly it's taking `controller interface{}` and using reflect to call a method.  Perhaps this was to break the `system` <--> `controllers` dependency cycle?  I'd much to prefer to pass a function with a certain signature, as in the [newly added APIHandler](https://github.com/decred/dcrstakepool/pull/73/files#diff-5467de60ded0da2fb4a582d6bfa29bddR148).  But that's for another day.

system/core.go
```diff
@@ -101,11 +110,8 @@ func (application *Application) Route(controller interface{}, route string) inte
 
 		body, code := method(c, r)
 
-		if session, exists := c.Env["Session"]; exists {
-			err := session.(*sessions.Session).Save(r, w)
-			if err != nil {
-				glog.Errorf("Can't save session: %v", err)
-			}
+		if err := saveSession(c, w, r); err != nil {
+			glog.Errorf("Can't save session: %v", err)
```
chappjc: And quit? or just let it ride?

chappjc:
Rebased, but not re-tested.

chappjc:
@jolan Aside from the rebase, these changes are fairly well tested and running in production, except the POST routes, which I disabled.

jolan:
Didn't look into it yet but running apiClientExample.sh (with the API version changed from 0.1 to 1) fails yet works on master.

chappjc:
@jolan Haven't tried the .sh script.  Do you have the error?

jolan:
I was receiving 'invalid session'.

chappjc:
Huh.  I'm actually not clear on how sessions are supposed to work with the API.  When I though about the API initially assumed the user would get an API key and they would use it.  But it seems like the current design requires a login + session through the browser.

jolan:
We discussed this on IRC but might as well put this here.  startsession is basically just a dummy call to trigger goji's middleware to set a session cookie and get a CSRF token so we can start posting.  The current API code is basically just the website logic returning JSON instead of HTML in response.

API keys are certainly a good idea as long as the implementation still allows a paymetheus user to easily sign up and register.

jolan:
Actually, I remember now that the current CSRF code wouldn't work well with session-less tokens:

https://github.com/decred/dcrstakepool/blob/master/system/middleware.go#L93-L126

chappjc:
@jolan But that middleware (ApplyCsrfProtection) isn't required for the API, is it?

jolan:
Considering we only allow one pubkeyaddr to be submitted, I think it's ok to drop CSRF for the API.

jrick:
I would still be concerned about CSRF between signing up for an account and submitting the pubkey.

controllers/main.go
```diff
@@ -138,54 +139,47 @@ func NewMainController(params *chaincfg.Params, adminIPs []string, baseURL strin
...
+			_, response, err = controller.APISignUp(c, r)
+		default:
+			return nil
 		}
 	}
 
-	return "{\"status\": \"error\"," +
-		"\"message\": \"invalid API command\"}\n", http.StatusOK
+	if err != nil {
+		status = err.Error()
```
jrick: in the current api, all errors have status "error", and the response is the error message itself.

chappjc: what's the purpose of the contents of `err` then?

jrick: it gets placed in the response data field.

that said this whole api is pretty wonky so i'm not defending its decisions.  error codes make a lot more sense than strings for status.

chappjc:
@jrick Guess I don't understand the purpose of CSRF.  This is to prevent 3rd parties from posting, rather than through/from a web page on the same domain.  But the point of the API _is_ public access.

After signup up, it's not a free for all to submit an address, there still must be a valid session with valid UserId.  This seems unrelated to CSRF.

chappjc:
Amended my last comment.  I think I convinced my self we shouldn't be doing session-based auth on the API.

controllers/main.go
chappjc: Alright, but the third output of API* (err) has no purpose other than to be nil or not.  The response field comes from the second output and is related but not equal to the contents of the error.

jrick: ¯\_(ツ)_/¯

jrick: it's not a massive issue in this case. the paymetheus implementation checks for success rather than checking for error, and assumes error if it's not success.

jrick:
I didn't consider the captcha issue before, but yeah I can understand that stake pool operators would be less inclined to support this API if it means robots could create accounts.  Maybe there needs to be code to remove accounts if no tickets are pushed for the voting address within some time frame.

chappjc:
How about Paymetheus directs you to the sign up page, where you get your API key, and the key is paired with an address in the settings.  Then the key is passed as a URL argument or in the request header.  HTTPS required.

jrick:
It is another hoop for users to go through, but if it means that more stake pool operators will support the API, probably a good idea.

jolan:
The overarching goal is to get more individuals staking by reducing the friction in Paymetheus so from that perspective I don't like adding another step. However, if you're not comfortable operating the pool like that, then people can't stake there anyway.

So I'm fine with requiring users to signup manually and generating an API key.

> and the key is paired with an address in the settings

Email address? If so, yes that sounds good to me.

> HTTPS required.

Sounds good to me.  Some of the pools are plain HTTP which neither @jrick or myself like so forcing the issue is fine too.

chappjc:
@jrick I would also think that users would prefer an API key from a security perspective.  I would not like to give my username and password to an app (desktop, web or otherwise).  Plus. it should be a familiar exercise when setting up accounts to work with third-party apps.

@jolan The address I meant was the URL base that is paired with the key.  Say you were on multiple pools, Paymetheus settings could have a URL base as the basic attribute for a pool's settings entry, plus the key/token for it.  Then a standard API would be assumed and the key used in whatever way.

jolan:
It seems doing:

```go
		if err := saveSession(c, w, r); err != nil {
			log.Errorf("Can't save session: %v", err)
		}

		apiResp := apiFun(c, r)

		if err := saveSession(c, w, r); err != nil {
			log.Errorf("Can't save session: %v", err)
		}
```

fixes it.

saveSession() needs to be called after the function is run to save the session modification that happens in APISignIn.

chappjc:
@jolan huh.  Good catch.  Does it only need to be after?  Like https://github.com/chappjc/dcrstakepool/blob/48b8868a8a03adced2c413e6a39073c981e336da/system/core.go#L111

jolan:
I tried just after but that didn't work for me for reasons I don't understand yet.  I'll play around a bit more tomorrow unless you beat me to it.

server.go
```diff
@@ -128,7 +134,8 @@ func main() {
 	app.Post("/address", application.Route(controller, "AddressPost"))
 
 	// API
-	app.Handle("/api/*", application.Route(controller, "API"))
+	app.Handle("/api/v1/:command", application.APIHandler(controller.API))
```
jolan: Can you change this back to v0.1 for now since it's compatible?

system/core.go
```diff
@@ -130,3 +135,54 @@ func (application *Application) Route(controller interface{}, route string) inte
...
+	}
+	return log.Errorf("Session not available")
+}
+
+func (application *Application) APIHandler(apiFun func(web.C, *http.Request) *APIResponse) web.HandlerFunc {
+	return func(c web.C, w http.ResponseWriter, r *http.Request) {
+		if err := saveSession(c, w, r); err != nil {
+			log.Errorf("Can't save session: %v", err)
+		}
+
```
jolan: I was either wrong or had some junk in my tree. Moving saveSession() to after apiFun() does works ok.

chappjc:
rebased on master so merge can be a ff

jolan:
Thank you! :+1:

chappjc:
:shipit:  :hammer:

chappjc:
@jolan & @jrick For the API auth stuff you're planning to work on, I'd just like to suggest [JWT (JSON Web Tokens)](https://jwt.io/introduction/) as a possible solution for API key generation and verification.  I has a number of nice properties.  Not the least awesome is the ability to encode any type of claims, such as expiration, permissions, etc.

I had forgotten about this until now, but I made a demo project https://github.com/chappjc/resttest based on a sample from Auth0.  There are a few good implementations of JWT for go, but I used https://github.com/dgrijalva/jwt-go.  The middleware is simple to write but Auth0 had a package.  Our own would be preferable though.

PR #74, opened 2016-12-05
Flush log before exit and wrap main().

chappjc:
Following up on PR https://github.com/decred/dcrstakepool/pull/67.
There's a need to flush the log backend before os.Exit().  Do like other projects and wrap main() so deferred functions get run.
Try using `--listen=[wrongIP]:8000` and you get no messages, it just exits.  You should see:  `could not bind listen tcp [wrongIP]:8000: bind: cannot assign requested address`


jolan:
Ah right.  Good catch.

PR #76, opened 2016-12-14
add API Tokens, change API version to v1 and bump to 1.0.0

### LINKED ISSUE #75: Improve API
JSON messages need to be updated to not use strings for literally all fields.

There has been discussion about redirecting users to the pool's web page in order to sign up for the service, since then it can be protected with a captcha.

Consider using a sessionless API key instead of logging in to the service.

Add error codes to report specific error conditions.

jolan:
This addresses everything mentioned in #75.

- [x] JSON messages need to be updated to not use strings for literally all fields.

- [x] There has been discussion about redirecting users to the pool's web page in order to sign up for the service, since then it can be protected with a captcha.

- [x] Consider using a sessionless API key instead of logging in to the service.

- [x] Add error codes to report specific error conditions.

@chappjc mentioned requiring TLS but since TLS termination can be handled by nginx, I'm not sure how that would work.  Also @jrick mentioned keeping CSRF but that requires sessions or a 'get csrf token' API call.  Not sure it is worth the effort.

Anyway, this PR works well for me. Just needs more testing and some review and close out the discussion items mentioned above.

Closes #75.


system/core.go
```diff
@@ -179,7 +186,8 @@ func WriteAPIResponse(resp *APIResponse, code int, w http.ResponseWriter) {
 // http.HanderFunc.
 func APIInvalidHandler(w http.ResponseWriter, _ *http.Request) {
 	resp := &APIResponse{Status: "error",
-		Message: "invalid API version",
+		Code:    codes.InvalidArgument,
```
jrick: I think this should be Unimplemented.

jrick:
I'll begin updating the client libs in Paymetheus to test the new API.

chappjc:
Won't be able to test or review for a while after all.  :(  Very exciting changes though!

poolapi/apijson.go
```diff
@@ -13,31 +13,32 @@ type Response struct {
 // TODO: make JSON tags lower-case and add "_" between words
 
 type PurchaseInfo struct {
-	PoolAddress   string `json:"PoolAddress"`
-	PoolFees      string `json:"PoolFees"`
-	Script        string `json:"Script"`
-	TicketAddress string `json:"TicketAddress"`
+	PoolAddress   string  `json:"PoolAddress"`
+	PoolFees      float64 `json:"PoolFees"`
```
jrick: Can this be changed to be a value between 0 and 1?  Easier to work with percentages this way.  I'd argue that percent formatting should only be done for display.

jrick: Actually I found my code to do the conversion client side.  If you think it makes sense to switch then please do so (and on the pool list at https://decred.org/api/?c=gsd) otherwise i'll continue using my converter.

jolan: Well that is designed so users can copy/paste the value for poolfees= in dcrwallet.conf/ticketbuyer.conf.  Not sure why it would need to be changed.

system/core.go
jolan: Well that's a catch all right now for invalid version OR invalid command so I think InvalidArgument makes more sense.

jolan:
@chappjc Well if there's anything major please let me know now.  We'll probably roll with this unless there are any blockers due to time pressure.

jrick:
Paymetheus client lib implemented, and new API tested and confirmed working: decred/Paymetheus#212

chappjc:
@jolan As much as I'd like to test it out, I don't have time.

So the Paymetheus in 0.7.0 will work like this, or is this planned for a later release?  I guess I'll reopen the pool if there is a chance for many new signups for across all the pools.  Somehow I doubt there will be many though.

jrick:
@chappjc unsure if I will have time to finish the integration with Paymetheus proper, but the client libs (`Paymetheus.StakePoolIntegration`) are updated for v1.  Been testing them in the C# REPL and are working fine.

system/middleware.go
```diff
@@ -41,6 +42,45 @@ func (application *Application) ApplyDbMap(c *web.C, h http.Handler) http.Handle
 	return http.HandlerFunc(fn)
 }
 
+func (application *Application) ApplyAPIAuth(c *web.C, h http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		if c.Env["IsAPI"].(bool) {
+			var apitoken string
+
+			authHeader := r.Header.Get("Authorization")
+			if strings.Index(authHeader, "Bearer ") == 0 {
```
chappjc: was going to suggest that too, but for readability

This is in a few spots I think.

jolan:
I'm not really sure how things will shake out but it'd be nice to get this PR merged and then have all the stake pools upgraded so the stage is set at least.

chappjc:
glide.yaml doesn't include dgrijalva/jwt-go but the lock does

system/middleware.go
```diff
@@ -41,6 +42,45 @@ func (application *Application) ApplyDbMap(c *web.C, h http.Handler) http.Handle
...
+						return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
+					}
+					return []byte(application.APISecret), nil
+				})
+
+				if err != nil {
+					log.Warnf("invalid token %v: %v", apitoken, err)
+				} else {
+					if claims, ok := JWTtoken.Claims.(jwt.MapClaims); ok && JWTtoken.Valid {
+						dbMap := c.Env["DbMap"].(*gorp.DbMap)
```
chappjc: These type assertions on a map with `interface{}` values make me nervous.   This will also panic if `c.Env["DbMap"]` was not set prior to this middleware.  A nil interface is not a nil pointer (value).
https://play.golang.org/p/9ncok9RicI

system/middleware.go
```diff
@@ -41,6 +42,45 @@ func (application *Application) ApplyDbMap(c *web.C, h http.Handler) http.Handle
 	return http.HandlerFunc(fn)
 }
 
+func (application *Application) ApplyAPIAuth(c *web.C, h http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		if c.Env["IsAPI"].(bool) {
+			var apitoken string
+
+			authHeader := r.Header.Get("Authorization")
+			if strings.HasPrefix(authHeader, "Bearer ") {
```
chappjc: perhaps `strings.HasPrefix(strings.ToLower(tokenString), "bearer ")` to be less strict (and the next line)?

system/middleware.go
```diff
@@ -41,6 +42,45 @@ func (application *Application) ApplyDbMap(c *web.C, h http.Handler) http.Handle
...
+					return []byte(application.APISecret), nil
+				})
+
+				if err != nil {
+					log.Warnf("invalid token %v: %v", apitoken, err)
+				} else {
+					if claims, ok := JWTtoken.Claims.(jwt.MapClaims); ok && JWTtoken.Valid {
+						dbMap := c.Env["DbMap"].(*gorp.DbMap)
+
+						user := models.GetUserById(dbMap, int64(claims["loggedInAs"].(float64)))
```
chappjc: `MapClaims` is a `map[string]interface{}` so all an attacker needs to do is pass a valid JWT (`jwt.Parse` is not hard to satisfy) and omit a `"loggedInAs"` claim.  The `float64` type assertion will panic.  No?

chappjc: The token we generate will have `"loggedInAs"`, but a client can send anything.

system/middleware.go
```diff
@@ -41,6 +42,45 @@ func (application *Application) ApplyDbMap(c *web.C, h http.Handler) http.Handle
 	return http.HandlerFunc(fn)
 }
 
+func (application *Application) ApplyAPIAuth(c *web.C, h http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		if c.Env["IsAPI"].(bool) {
+			var apitoken string
```
chappjc: scope unnecessarily broad.  `apitoken` is not used outside of following `if` statement, so I would just omit this line.

system/middleware.go
```diff
@@ -98,6 +150,11 @@ func strInSlice(strs []string, str string) int {
 
 func (application *Application) ApplyCsrfProtection(c *web.C, h http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
+		// disable CSRF for API requests
+		if c.Env["IsAPI"].(bool) {
```
chappjc: Like other middleware, this one has to come after ApplyIsAPI, so I think it should be noted and the potential panic handled.

chappjc:
https://github.com/decred/dcrstakepool/pull/76/commits/0dea7f8f13fccc762674632f971bacce40598e44 broke the lock hash and the yaml is still missing the jwt depend.  :)

chappjc:
Realize I was wrong about being easy to pass jwt.Parse given it checks the sig with the secret key, so it's not quite so easy to make it panic as I though.

system/middleware.go
```diff
@@ -41,6 +42,45 @@ func (application *Application) ApplyDbMap(c *web.C, h http.Handler) http.Handle
 	return http.HandlerFunc(fn)
 }
 
+func (application *Application) ApplyAPIAuth(c *web.C, h http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		if c.Env["IsAPI"].(bool) {
```
jolan: I merged them.

system/middleware.go
jolan: Fixed.

system/middleware.go
jolan: I looked into this a bit and it should be capitalized.

system/middleware.go
jolan: Well I am following the pattern that the rest of the app uses and it does that because it is set previously.

system/middleware.go
jolan: Yes the client can send anything but then it's an invalid non-signed token so that code isn't reached.

system/middleware.go
jolan: Well I added a nil check but again it should be set.  I do agree this could be cleaned up a lot but again I am mostly following the pattern the app has always used which is to execute the middleware in sequence and have the environment properly set up.  Not an unreasonable assumption to make in my opinion.

system/middleware.go
chappjc: Yeah, I don't know why it didn't bother me before.  Maybe because there are more middleware with a required order.  Whatever the reason I think there should be a note about the required order somewhere

system/middleware.go
chappjc: I cool with a nil check here.  This is the spot to give a message about what went wrong (middleware order)

system/middleware.go
chappjc: OK, I saw it like this in JWT.Parse so I thought it might have been a common thing

jolan:
I fixed some grammar in the README and addressed the review items as best I could in the time available.  Definitely more can be done in terms of cleanup and logging in the future.

chappjc:
`glide i` still says:

    [WARN]  Lock file may be out of date. Hash check of YAML failed. You may need to run 'update'

PR #79, opened 2017-02-02
return early with 0 tickets so we don't panic in SetTicketsVoteBits()

### LINKED ISSUE #78: panic in SetTicketsVoteBits() via TicketsPost() when user has 0 live tickets
Reported by @dyrk:

[9:22] if user has at least 1 tickets and 0 Live tickets
[9:22] he can still update votebits from Tickets page
[9:23]  and dcrstakepool dies

I believe this happens because the 'Update Tickets' button is shown when there's an immature ticket but since it's not actually live the code ends up calling SetTicketBoteBits() with no live ticket hashes.

jolan:
Closes #78.

Should probably also prevent the button from showing but that can be done as part of the big votebits handling rewrite that is coming soon.

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -1563,7 +1563,7 @@ func (controller *MainController) Tickets(c web.C, r *http.Request) (string, int
 	// spui := new(dcrjson.StakePoolUserInfoResult)
 	spui, err := w.StakePoolUserInfo(multisig)
 	if err != nil {
-		// Render page with messgae to try again later
+		// Render page with message to try again later
 		log.Infof("RPC StakePoolUserInfo failed: %v", err)
 		session.AddFlash("Unable to retreive stake pool user info.", "main")
 		c.Env["Flash"] = session.Flashes("main")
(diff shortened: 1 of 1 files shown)


chappjc:
What line was the panic if `len(liveTicketHashes)==0` ?  I'm not seeing it.

`liveTicketHashes` should include immature, so I don't follow how 0 mature and N immature will cause a panic.

jolan:
The output @Dyrk sent me was:

16:27:19 2017-02-02 [INF] CNTL: Started setting of vote bits for 0 tickets.
panic: runtime error: index out of range

goroutine 69 [running]:
panic(0x5912c0, 0xc42000c120)
    /usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/decred/dcrstakepool/controllers.(*walletSvrManager).SetTicketsVoteBits(0xc420529900, 0x0, 0x0, 0x0, 0xa966d8, 0x0, 0x0, 0x0, 0x0)
    /Users/dyrk/go/src/github.com/decred/dcrstakepool/controllers/dcrclient.go:1311 +0x50b
github.com/decred/dcrstakepool/controllers.(*MainController).TicketsPost.func1(0xc420529900, 0xa2c8c0, 0xc420546500, 0x0, 0xc4204a8fc0)
    /Users/dyrk/go/src/github.com/decred/dcrstakepool/controllers/main.go:2336 +0x350
created by github.com/decred/dcrstakepool/controllers.(*MainController).TicketsPost
    /Users/dyrk/go/src/github.com/decred/dcrstakepool/controllers/main.go:2349 +0x666

So it would get down to SetTicketVoteBits() with 0 hashes which could cause the panic.  I couldn't reproduce locally so maybe the conditions are different than what was reported.  In any event, a simple protective fix that doesn't hurt anything.

PR #81, opened 2017-02-04
tplstakepool - New Design for dcrstakepool

### LINKED ISSUE #80: Design Update, New Template, integrate new HTML, CSS and JS
We need to Update the default Design Template on dcrstakepool.

I will make a design update and integrate new HTML, CSS and JS

karamble:
This is the initial commit for the new design

Closes #80 

It will create a new directory /tplstakepool into the project.

Files including:
`views/`
Includes HTML pages and subpages with logic and all {{.variables}} to display the dcrstakepool content

`public/`
Includes CSS, JS and Fonts

`src/
bower.json
Gruntfile.js
package.json`
If you want to customize the css and js files, edit the files in directory src/ accordingly. Afterwards you be able to render a optimized and minified version into the public/ directory. See included Readme.md on this steps.

Or read them here:

### Step 1 (Optional, when customizing css):
```
$ cd $GOPATH/src/github.com/decred/dcrstakepool/tplstakepool
$ npm install
$ bower install
$ grunt
```
### Step 2
Edit your dcrstakepool.conf  and point to the public and views directory to activate the template
```
publicpath=tplstakepool/public
templatepath=tplstakepool/views
```
### Step 3
Restart dcrstakepool


chappjc:
Looks cool, but the readme could use a couple of tips/details.

- Had to install bower.  Didn't want to do it globally so I used npm install bower, and called it from ./node_modules/.bin/bower.  I don't even know if this is right, but it seemed to work.
- grunt errored about Ruby and Sass not installed in my PATH.

chappjc:
Maybe also update .gitignore with build artifacts.  After following the optional instructions (step 1), I ended up with the following

```
[jon@arch-monkey-envy] 10:27 $ git status
On branch templateredesign
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.sass-cache/
	bower_components/
	node_modules/
	public/index.html
	public/signin.html
	public/stats.html
	public/submit.html
```

I'm not sure what the .html files are doing there though.  Adding the three folders above seems like a good idea though.

karamble:
Yeah i will add somethings about the sass-css+javascript rendering-chain in the readme.

What happens there is that all needed Javascript Libraries will get pulled from their respective repositories. Then they will be joined, rendered and minified into single files. The output of this process will get into the /public directorys.

This PR comes with those /public assets (JS,CSS,FONTS) prerendered in the /public directory. The user should be able to rebuild those files without the need that we include all those 3rd partie libraries into the repo.

This process makes shure that the user is able to customize the design, download all needed JS,CSS,FONTS files from official repositories and is able to update them independent from the repo.

Those three directorys you have mentioned above (node_modules, bower_components, .sass-cache) should deff. not go into the PR as those are local cache files, needed node and bower plugins to do these javascript, merging and minify job. These can be easily deleted after the building building process which renders files directly into the /public directory.

karamble:
Ok will do as you suggested. The Logo simply wasn't there. It was shared yesterday on the forums the 1st time. Thanks for your feedback, appreciated.

jolan:
There's some changes I'm not so sure about (especially the tickets page), but no functional issues were encountered and it's easy enough to change things back if it is determined best to do so.

PR #83, opened 2017-02-05
Compute missed excluding expired in the template

chappjc:
This is just for discussion, since it it's undecided how missed should be displayed.  Plus it brings in a dependency, but it's shorter than writing the function ourselves.

This math is done in the template to facilitate operators changing the semantics of their stats without recompiling golang.  When/if dcrwallet changes, this is easy to change back.

### DIFF
--- a/system/core.go
+++ b/system/core.go
@@ -11,6 +11,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/Masterminds/sprig"
 	"github.com/decred/dcrstakepool/codes"
 	"github.com/decred/dcrstakepool/models"
 	"github.com/gorilla/sessions"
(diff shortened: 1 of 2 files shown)


karamble:
Yes i think we should substract the expired from the missed tickets.

Some of the Pools allready did this by theirselfs so currently they are cheating their stats, it should be rolled out to every pool as it is the correct value users are expecting.

chappjc:
I'd update glide, but something is wrong with gpkg.in and glide:

```
[INFO]	--> Fetching gopkg.in/gorp.v1
[WARN]	Unable to checkout gopkg.in/gorp.v1
[ERROR]	Error looking for gopkg.in/gorp.v1: Unable to get repository
```

Can't figure it out right now.

Besides, @karamble points out that proportion missed also needs to be computed.

chappjc:
Oh, yeah, the API.  This only handles the web interface.  I guess it's necessary to change this in the go code.

PR #86, opened 2017-02-14
Add options for Hard Fork Test Demo

dajohi:


jolan:
This needs to be wrapped in checks for testnet so the behavior on mainnet isn't modified.

dajohi:
Good call :)

Updated...

Issue #84, opened 2017-02-13
should use SplitHostPort in getClientIP()


jolan:
We're splitting on : which more than likely breaks IPv6. There's a convenience function for this:

https://godoc.org/net#SplitHostPort

PR #102, opened 2017-02-27
Use net.SplitHostPort instead of string splitting.

chappjc:
Closes https://github.com/decred/dcrstakepool/issues/84
go fmt also alphabetized imports

NOT yet tested.

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -82,12 +82,11 @@ func randToken() string {
 // real_ip module to correctly set the X-Real-IP header.
 func getClientIP(r *http.Request, realIPHeader string) string {
 	// getHost returns the host portion of a string containing either a
-	// host:port formatted name or just a host. An empty string is returned if
-	// net.SplitHostPort returns an error.
+	// host:port formatted name or just a host.
 	getHost := func(hostPort string) string {
 		ip, _, err := net.SplitHostPort(hostPort)
 		if err != nil {
-			return ""
+			return hostPort
 		}
 		return ip
 	}
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -61,6 +61,7 @@ type MainController struct {
 	poolLink             string
 	params               *chaincfg.Params
 	rpcServers           *walletSvrManager
+	realIPHeader         string
 	recaptchaSecret      string
 	recaptchaSiteKey     string
 	smtpFrom             string
@@ -79,7 +80,7 @@ func randToken() string {
 // Get the client's real IP address using the X-Real-IP header, or if that is
 // empty, http.Request.RemoteAddr. See the sample nginx.conf for using the
 // real_ip module to correctly set the X-Real-IP header.
-func getClientIP(r *http.Request) string {
+func getClientIP(r *http.Request, realIPHeader string) string {
 	// getHost returns the host portion of a string containing either a
 	// host:port formatted name or just a host. An empty string is returned if
 	// net.SplitHostPort returns an error.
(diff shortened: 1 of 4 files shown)


controllers/main.go
```diff
@@ -79,10 +80,13 @@ func randToken() string {
...
-		if strings.Contains(ip, ":") {
-			parts := strings.Split(ip, ":")
-			return parts[0]
+	// getHost returns the host portion of a string containing either a
+	// host:port formatted name or just a host. An empty string is returned if
+	// net.SplitHostPort returns an error.
+	getHost := func(hostPort string) string {
+		ip, _, err := net.SplitHostPort(hostPort)
+		if err != nil {
+			return ""
```
jrick: why not return the input?

chappjc: if it's not a valid host in the eyes of `net.SplitHostPort`, is it right to use it?

jrick: depends, where is the input coming from?

chappjc: In this case there are two possible sources: (1) whatever is in the `X-Real-IP` header or (2) the `RemoteAddr` field of the request as a backup.

The first is set by the proxy/load-balancer/whatever and 2 comes from golang's net library and is presumably set to a valid value, probably whatever interface it's bound to or server the request from.

jrick: why do we even care about that first value? I think we should only be looking at the client's remote address in the request itself, not a header.

chappjc: https://github.com/decred/dcrstakepool/issues/63

jolan: @jrick Typically when you use nginx as the frontend and have it proxy to the app, remoteaddr will always be 127.0.0.1. Thus most people configure nginx to put the real IP in that X-Real-IP header.

jrick: hmm ok I see. I think that code should only be enabled when a proxy is being used, otherwise only the request remote address is used.

chappjc: @jrick In that case, `X-Real-IP` would be empty and would default to `RemoteAddr`.

jrick: couldn't it be set by the remote client?

chappjc: yeah, I don't know.  I guess we need to answer that and/or make room for the possibility of running without a reverse proxy.

jrick: how about checking the request remote address first and if it is a loopback address (::1 or 127.0.0.1/8) then use the header?

chappjc: It can be any number of other non-routable IPs and/or proxies.  The other approach, which was implemented by @jolan is to look at the IPs listed in X-Forwarded-For to find one that is not a proxy or non-routable.  That is exactly what [ngx_http_realip_module](http://nginx.org/en/docs/http/ngx_http_realip_module.html) is designed to do robustly under the hood.

chappjc: For example, http://serverfault.com/a/414166/190378

jrick: probably makes sense then to specify the ip of the proxy in the config and only fallback to the header when it matches.

chappjc: If someone is proxied and they set it up right, there's no need to check anything as there's no possibility for the client to spoof the header - it's set by the proxy.  Sounds like we want to handle the case of someone who's not proxied so that they don't get a spoofed X-Real-IP.  Assuming we want to support running w/o a proxy, then all that's needed is a boolean flag.

jrick: I don't want to assume that every proxy behaves the same and that each use the same headers.  How about we set the header to use in the config and if unset, it defaults to the request addr?

chappjc: @jolan Did you solution avoid this problem?  I mean did it step through the addresses in X-Forwarded for from right to left so as to get the first untrusted IP instead of the last, which could be spoofed?

chappjc: @jolan That's a reasonably clean solution.  No objection to that.

jolan: @chappjc Yes it went from right to left but I had it checking both X-Forwarded-For and X-Real-IP:

https://github.com/decred/dcrstakepool/compare/master...jolan:jolan_fixip

It can probably be cleaned up a bit to use net.IPNet/ipNet.Contains(ip).

chappjc:
With https://github.com/decred/dcrstakepool/pull/102/commits/107368f25ad3b57bed7ee2a3634cf828bb2e5062 the default is to use `RemoteAddr` otherwise `realipheader` string from the config.  Which reminds me, the sample config needs the option added.

dajohi:
Rebase needed.

chappjc:
Rebased.

Latest changes pertaining to real-ip discussion need verification.

chappjc:
Fixed handling just a host with no port in header.  I must have misunderstood semantics of SplitHostPort.

Tested on testnet.

Issue #99, opened 2017-02-27
creating voting daemon to takeover stakepool-related functionality from dcrwallet


jolan:
Broken out from #49.

- [ ] create a voting daemon (tentatively named stakepoold) to handle the voting
- [ ] receive voting config for all users from dcrstakepool
- [ ] set votebits according to user preference, ask wallet to sign, broadcast

PR #109, opened 2017-03-13
remove deprecated fields and disable stakemining safety check

### LINKED ISSUE #108: remove stakemining check
stakemining is deprecated and being removed so the stakemining safety check needs to be adjusted/updated.

jolan:
Closes #108.

Related to #99.

Basically where we're at now:

- dcrwallet 0.8.2 has deprecated enablestakemining
- dcrd master has various walletinfo fields removed
- dcrwallet has a pr pending to update the walletinfo rpcs
- We're moving to a voting daemon for 1.0 rather than having the wallet vote directly.

So with all that being taken into account, I think it's best to just remove the stakemining safety check.  The pools are well established now so hopefully it won't cause an issue to remove this minimal safety check.

### DIFF
--- a/controllers/dcrclient.go
+++ b/controllers/dcrclient.go
@@ -1524,9 +1524,6 @@ func (w *walletSvrManager) CheckServers() error {
 		if !wi.DaemonConnected {
 			return fmt.Errorf("Wallet on svr %d not connected\n", i)
 		}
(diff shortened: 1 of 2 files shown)


chappjc:
Looks good.  I just commented the stakemining checks after the walletinfo changes anyway.

PR #112, opened 2017-03-27
Refactored markup

captain-redbeard:
Refactored markup of all views improving readability, removing redundant code and fixing minor issues.


jolan:
Just wanted to let you know that I'm aware of this but we're still shaking out bugs from the last re-design so I'm going to be overly cautious with this and it may have to sit for awhile until myself or @karamble can properly review it.

It'd help a lot if you could minimize the diff by not re-indenting everything so it's easier to review.

captain-redbeard:
The actual changes are very minimal, for e.g. removing span classes that are not required and fixing some css.

The major change is the indenting and refactoring to improve readability.

captain-redbeard:
Expanding on the changes to hopefully speed up the review.

- refactored html markup in every view
- removed non required spans
- fixed footer width
- fixed input field widths
- split css into separate files for easier editing
  - bootstrap.css
  - styles.css
- added theme-color to header.html and maintenance.html
- updated instructions in tickets.html.

PR #115, opened 2017-03-28
Always allow registered users to login and submit their address

### LINKED ISSUE #114: Registered users cannot use their API key (PM) if the pool is closed to signups
Closing the pool is only supposed to apply to signups

chappjc:
Closing the pool should only prevent new signups, but presently it prevents registered users from logging in and uploading their address.

Remove the close pool checks from the API address handler, the /address POST handler, and the signin POST handler.

SignUp is still restricted when closePool is true.

Work in progress in that I will update when done testing.

Closes #114.

### DIFF
--- a/config.go
+++ b/config.go
@@ -393,7 +393,7 @@ func loadConfig() (*config, []string, error) {
 	activeNetParams = &mainNetParams
 	if cfg.TestNet {
 		numNets++
-		activeNetParams = &testNetParams
+		activeNetParams = &testNet2Params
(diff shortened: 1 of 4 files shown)


chappjc:
Addresses https://github.com/decred/dcrstakepool/issues/114.  (I know, I forgot to mention the issue in the commit).

chappjc:
Huh, looks like I need to make a testnet**2** pool now.  phooey

PR #119, opened 2017-04-03
gosimple and other lint fixes

chappjc:
Make travis happy/happier by addressing gosimple issues.

- It is unnecessary to specify both the size and capacity of a slice when they are equal values.
- Document ErrSetVoteBitsCoolDown and NewGetTicketsCacheData.


chappjc:
Travis is finally happy.

PR #120, opened 2017-04-04
Markup changes

captain-redbeard:
Split fixes into separate commits for easier reviewing and does not include refactor.
Each commit has the detailed description of the change.

_Note; the major change is splitting main.css into two files, bootstrap.css and styles.css which was done with this online tool: http://unminify.com/
This is to allowing for easier editing of the styles._

Let me know if anything else is required to perform this PR!


chappjc:
Especially glad to see the ticket instructions updated in https://github.com/decred/dcrstakepool/pull/120/commits/45ca50f88c9ee32a4eba177515105cf4884b1818

@jolan Do you know of the ticket buyer settings (e.g. `ticketbuyer.maxpriceabsolute`) need to be under the `[Ticket Buyer Options]` header?  I thought I saw something about that on slack, but maybe it was just the `ticketbuyer.` prefix and the header is ignored...

PR #121, opened 2017-04-04
Replaced hard coded min server value with config value

captain-redbeard:
As per title and commit description.

### DIFF
--- a/config.go
+++ b/config.go
@@ -526,7 +526,7 @@ func loadConfig() (*config, []string, error) {
 	// Add default wallet port for the active network if there's no port specified
 	cfg.WalletHosts = normalizeAddresses(cfg.WalletHosts, activeNetParams.WalletRPCServerPort)
 
-	if len(cfg.WalletHosts) < 2 {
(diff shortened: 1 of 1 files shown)


jolan:
defaultMinServers is for wallets **connected** not wallets **configured**.  And I kind of disagree with making it a constant since that sort of implies it could be easily changed to a different value.  A value of 1 will blow up since the code assumes 2+ everywhere.  Maybe a comment explaining why it's hard coded is best.

config.go
```diff
@@ -526,7 +526,7 @@ func loadConfig() (*config, []string, error) {
 	// Add default wallet port for the active network if there's no port specified
 	cfg.WalletHosts = normalizeAddresses(cfg.WalletHosts, activeNetParams.WalletRPCServerPort)
 
-	if len(cfg.WalletHosts) < 2 {
+	if len(cfg.WalletHosts) < defaultMinServers {
```
jolan: Per previous comment, I think just explaining why it's hardcoded at 2 makes more sense.

PR #122, opened 2017-04-04
Conforming to mysql naming conventions

captain-redbeard:
Conforming to mysql naming conventions: https://dev.mysql.com/doc/internals/en/naming-conventions.html
- table names updated
- column names updated
- change COUNT(*) to COUNT(x_id)
- migrate sql script created

**_Note; existing pools must run the migrate script to update the existing database table and column names._**


chappjc:
You can do

    go fmt `glide nv`

at the root of the repo.

jolan:
I'm not sure this is enough of a benefit to warrant adding a manual intervention step to the upgrade process.  If anything, we want to switch to a better ORM that is more DB-agnostic in which case we wouldn't care about MySQL naming conventions.

Maybe some of the stake pool operators will chime in.

@chappjc @girino @karamble @jyap808 @mverrilli @dyrk @418sec @Retord @walkeralencar @tpruvot

girino:
I agree with @jolan. unless the changes bring some real benefit for the pool users or maintainers, i'd rather not add any manual upgrade step. Database migration in production systems are a pain in the ass.

captain-redbeard:
The only benefit in the changes is changing COUNT in the queries which can be done with or without the name changes.

With that said, I still believe it would be better to adopt the standard conventions sooner rather than later as there may become more pools and more information to be stored in the database. 

Either way, not a huge issue but it does make it easier for new devs to jump straight in.

jyap808:
Er, yeah I don't want to change anything. This doesn't have an obvious benefit.

Issue #111, opened 2017-03-24
Old references to `enablestakemining`


infertux:
There are references to `enablestakemining` in _sample-dcrwallet.conf_ and _views/tickets.html_ which is no longer supported since https://github.com/decred/dcrwallet/pull/606.

jolan:
Things got a bit messy because enablestakemining was deprecated in 0.8.2 but the stakepool code at the time of the 0.8.2 release still had a safety check to make sure that was enabled.

Anyway, I was aware of the discrepancy and it will get fixed soon for the next release.  It's nice to have a tracking issue on it though, thank you.

PR #127, opened 2017-04-17
various cleanups

dajohi:
- enablestakemining -> enablevoting in samepl-dcrwallet.conf
- remove unused functions
- use fileExists API


dajohi:
Fixes #111

PR #138, opened 2017-04-19
Send votes to dcrd

dajohi:


dajohi:
Try again with latest.  Paste trace.

PR #140, opened 2017-04-20
backend: Fix computing 1 extra fee address, add test

chappjc:
`deriveChildAddresses` takes a start index and a count, not a [start, end) range.

Add server_test.go, and test for `calculateFeeAddresses` using a testnet extended public key.

### DIFF
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -110,10 +110,9 @@ func calculateFeeAddresses(xpubStr string, params *chaincfg.Params) (map[string]
 
 func deriveChildAddresses(key *hdkeychain.ExtendedKey, startIndex, count uint32, params *chaincfg.Params) ([]dcrutil.Address, error) {
 	addresses := make([]dcrutil.Address, 0, count)
-	for i := uint32(0); i < count; i++ {
+	for i := uint32(0); i < count; {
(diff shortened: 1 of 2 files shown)


backend/stakepoold/server.go
```diff
@@ -112,6 +113,7 @@ func deriveChildAddresses(key *hdkeychain.ExtendedKey, startIndex, count uint32,
 	for i := uint32(0); i < count; i++ {
 		child, err := key.Child(startIndex + i)
 		if err == hdkeychain.ErrInvalidChild {
+			// NOTE: return slice will be shorter than requested
```
jolan: We should probably just error out if we hit this but I don't see why it would happen in practice.

chappjc: I'm really not clear on why this happens at all so whatever error you suggest I will put in there

jrick: https://godoc.org/github.com/decred/dcrutil/hdkeychain#ExtendedKey.Child

chappjc: Fair enough.  Still doesn't explain to me why their is that remote possibility, but that's OK.

chappjc: Oops, I pushed the fix on the wrong branch.  Putting it here too.

PR #144, opened 2017-04-20
start adding a test framework

marcopeereboom:
WIP don't look

### DIFF
--- a/backend/stakepoold/rpcclient.go
+++ b/backend/stakepoold/rpcclient.go
@@ -173,8 +173,8 @@ func walletSendVote(ctx *appContext, hexTx string) (*chainhash.Hash, error) {
 	return ctx.walletConnection.SendRawTransaction(newTx, false)
 }
 
-func walletFetchUserTickets(ctx *appContext) map[string]UserTickets {
-	userTickets := map[string]UserTickets{}
+func walletFetchUserTickets(ctx *appContext) map[string]string {
+	userTickets := make(map[string]string)
 
 	ticketcount := 0
 	usercount := 0
@@ -192,27 +192,20 @@ func walletFetchUserTickets(ctx *appContext) map[string]UserTickets {
 			continue
 		}
 		if spui != nil && len(spui.Tickets) > 0 {
-			tickets := make([]*chainhash.Hash, 0)
 			for _, ticket := range spui.Tickets {
 				switch ticket.Status {
 				case "live":
 					hash, err := chainhash.NewHashFromStr(ticket.Ticket)
 					if err != nil {
 						log.Infof("invalid ticket %v", err)
-					} else {
-						tickets = append(tickets, hash)
+						continue
 					}
+
+					userTickets[hash.String()] = ctx.userVotingConfig[msa].MultiSigAddress
+					ticketcount++
 				}
 			}
-			if len(tickets) > 0 {
-				userTickets[msa] = UserTickets{
-					Userid:          ctx.userVotingConfig[msa].Userid,
-					MultiSigAddress: ctx.userVotingConfig[msa].MultiSigAddress,
-					Tickets:         tickets,
-				}
-				usercount++
-				ticketcount = ticketcount + len(tickets)
-			}
+			usercount++
 		}
 	}
 
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -1,3 +1,7 @@
+// Copyright (c) 2017 The Decred developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
 package main
 
 import (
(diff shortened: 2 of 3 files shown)


chappjc:
I was actually just adding benchmarks for `processWinningTickets`, using the actual pool from mainnet 126316 and the real winning tickets.  I guess I'll stop doing this since you're on it.

But in general, using `sliceContains` is no good for this.  It's O(N), while a map would be O(log N).  Thus I think a way to proceed is by changing the user's `Tickets` into a map instead of a slice.

marcopeereboom:
I am adding a map and benchmarks to prove the theory.

chappjc:
Cool, I'll step back from this then so I don't step on your toes.

marcopeereboom:
I'll look at your diff though; it may be moar cleverder!

backend/stakepoold/server_test.go
```diff
@@ -30,28 +33,39 @@ var (
...
+		ut.Tickets = append(ut.Tickets, ticket)
+		c.userTicketsOld[msa] = ut
+
+		// NEW: use ticket as the key
+		tS := ticket.String()
+		utNew := c.userTicketsNew[tS]
+		utNew.Userid = int64(uid)
+		utNew.MultiSigAddress = msa
+		utNew.Tickets = append(utNew.Tickets, ticket)
+		c.userTicketsNew[tS] = utNew
```
chappjc: Switching `c.userTicketsNew` to a map with tickets as the key and then embedding arrays of tickets doesn't seem right, or maybe I'm just confused.

I was thinking to leave the multisig-keyed map, but change the definition of `UserTickets` to:

```golang
type UserTickets struct {
	Userid          int64
	MultiSigAddress string
	Tickets         map[*chainhash.Hash]struct{}
}
```

That way you can just replace

```if !sliceContains(ctx.userTickets[msa].Tickets, ticket) {```

with

```if _, ok := ctx.userTickets[msa].Tickets[ticket]; !ok {```

File: backend/stakepoold/rpcclient.go [hunk 2]
```diff
@@ -224,3 +224,50 @@ func walletFetchUserTickets(ctx *appContext) map[string]UserTickets {
...
+		}
+		if spui != nil && len(spui.Tickets) > 0 {
+			for _, ticket := range spui.Tickets {
+				switch ticket.Status {
+				case "live":
+					hash, err := chainhash.NewHashFromStr(ticket.Ticket)
+					if err != nil {
+						log.Infof("invalid ticket %v", err)
+						continue
+					}
```
jolan: We should just skip this validation.  If wallet is handing out invalid ticket hashes then there are big wallet problems going on.  And the ticket hashes won't get called to vote anyway.

backend/stakepoold/rpcclient.go
```diff
@@ -224,3 +224,50 @@ func walletFetchUserTickets(ctx *appContext) map[string]UserTickets {
...
+
+					userTickets[hash.String()] = UserTickets{
```
jolan: Then just use ticket.Ticket here.

backend/stakepoold/rpcclient.go
marcopeereboom: I want it logged.  Shouldn't happen anyway.

backend/stakepoold/rpcclient.go
marcopeereboom: ./rpcclient.go:204: cannot use ticket (type dcrjson.PoolUserTicket) as type string in map index

backend/stakepoold/rpcclient.go
```diff
@@ -192,27 +192,20 @@ func walletFetchUserTickets(ctx *appContext) map[string]UserTickets {
...
 				case "live":
 					hash, err := chainhash.NewHashFromStr(ticket.Ticket)
 					if err != nil {
 						log.Infof("invalid ticket %v", err)
-					} else {
-						tickets = append(tickets, hash)
+						continue
 					}
+
+					userTickets[hash.String()] = ctx.userVotingConfig[msa].MultiSigAddress
```
chappjc: You just did `hash, err := chainhash.NewHashFromStr(ticket.Ticket)`.  So why do `userTickets[hash.String()]` when you already have the string?  Isn't `userTickets[ticket.Ticket]` correct?

backend/stakepoold/server.go
```diff
@@ -381,45 +380,65 @@ func (ctx *appContext) loadData() {
...
+
+func (ctx *appContext) processWinningTickets(wt WinningTicketsForBlock) {
+	work := false
 	txStrs := make([]string, 0, len(wt.winningTickets))
 	for _, ticket := range wt.winningTickets {
-		txStrs = append(txStrs, ticket.String())
-		for msa := range ctx.userTickets {
-			if !sliceContains(ctx.userTickets[msa].Tickets, ticket) {
-				continue
+		tS := ticket.String()
```
chappjc: This would again would be unnecessary if `ctx.tickets` were a `map[chainhash.Hash]string` instead of a `map[string]string`.

backend/stakepoold/rpcclient.go
chappjc: But my general observation is that the map could just be a `map[chainhash.Hash]string` instead to speed up checking the winning tickets.  See other my comments please.

marcopeereboom:
Yeah I can do that.  Or better, do you have a pr?

chappjc:
Hah, no, sorry.  But the performance seems pretty damn good already, so...

PR #145, opened 2017-04-20
Log number of votes and amount of time to send them.

dajohi:
While here, small code cleanups.


marcopeereboom:
I already did that.

Issue #134, opened 2017-04-19
switch stakepoold to obtain user voting preferences via gRPC


jolan:
stakepoold currently gets the user config directly from MySQL which isn't ideal.

We could send the data via gRPC rather than triggering stakepoold to refresh like we do now.

PR #149, opened 2017-04-21
use gRPC to trigger stakepoold to update voting config

jolan:
Addresses 2 items from #134.  Pretty rough but does function as described.


chappjc:
I'm not too excited about the need for MySQL servers on each stakepoold machine.  This is the requirement with this PR, right?  Some type of data storage is needed to keep the user voting config cached on each machine, but I'm wishing for simple file-based method.  Access will never be performed on a critical path, so it is reasonable to have a lighter weight solution than MySQL.

For example, I'm trying asdine/storm (BoldDB ORM) for data management in hardforkdemo and I think it might be a good application here.  So, In `package userdata`, I'd define:

```go
type UserVotingConfig struct {
	Userid          int64     `storm:"id,increment"`
	MultiSigAddress string    `storm:"unique"`
	VoteBits        uint16
	VoteBitsVersion uint32    `storm:"index"`
}
```

I can try this and make a follow-up PR if it works.

marcopeereboom:
@chappjc we can debate this post 1.0.0.  The Thing to realize is that we need a distributed storage of sorts.  Mysql is a pretty low bar to clear and it can mirror etc for redundancy.

jolan:
Well there are two concerns:

1) The data source for the user voting config
2) Voting according to user preference in case that data source is down/broken/deleted/whatever disaster occurred

Everything could be done by dcrstakepool sending the data via gRPC and a json file on disk (like dcrd does with peers.json).  But the gRPC parts would need to be built out a bit more to i.e. handle re-sending commands when the connection is unavailable and an update couldn't be sent.

I don't see a reason to put a MySQL server on each stakepoold machine.  Just use the one(s) that dcrstakepool talks to. If MySQL is down, then dcrstakepool is down and there's no updates until MySQL and dcrstakepool are back up.

I'm not sure it makes sense to send user rows from MySQL over gRPC when stakepoold can just talk to MySQL itself directly.

backend/stakepoold/userdata/userdata.go
```diff
@@ -0,0 +1,126 @@
...
+type DBConfig struct {
+	DBHost     string
+	DBName     string
+	DBPassword string
+	DBPort     string
+	DBUser     string
+}
+
+// UserData stores the current snapshot of the user voting config.
+type UserData struct {
```
marcopeereboom: This lock is guaranteed wrong.  This needs to be in appcontext because we need to serialize pointer access; not the data fields themselves.

backend/stakepoold/userdata/userdata.go
```diff
@@ -0,0 +1,126 @@
...
+// DBSetConfig sets the database configuration.
+func (u *UserData) DBSetConfig(DBUser string, DBPassword string, DBHost string, DBPort string, DBName string) {
+	dbconfig := &DBConfig{
+		DBHost:     DBHost,
+		DBName:     DBName,
+		DBPassword: DBPassword,
+		DBPort:     DBPort,
+		DBUser:     DBUser,
+	}
+	u.Lock()
```
marcopeereboom: see previous; doesn't lock access.

backend/stakepoold/userdata/userdata.go
```diff
@@ -0,0 +1,126 @@
...
+	u.DBConfig = dbconfig
+	u.Unlock()
+}
+
+// Get returns the current user data.
+func (u *UserData) Get() map[string]UserVotingConfig {
+	u.RLock()
+	data := u.UserVotingConfig
```
marcopeereboom: Same.

backend/stakepoold/userdata/userdata.go
```diff
@@ -0,0 +1,126 @@
...
+
+// Update updates the current user data.
+func (u *UserData) Update() error {
+	newData := u.MySQLFetchUserVotingConfig()
+	if newData == nil {
+		return errors.New("unable to fetch user data")
+	}
+
+	u.Lock()
+	u.UserVotingConfig = newData
```
marcopeereboom: one more

PR #154, opened 2017-04-27
fix previous by adding back registration for OnWinningTickets

jolan:


chappjc:
Big ok here!

jolan:
Travis 1.7 job seems broken and I can't restart it for some reason.  Oh well.

Issue #132, opened 2017-04-19
optimize stakepoold's voting process


jolan:
There are several TODOs in the code which may be duplicated below:

- [x] speed up winningtickets to user ticket list comparison
- [x] relay the vote to dcrd, not dcrwallet
- [x] ~~only refresh ticket lists periodically~~ this is now done without blocking voting
- [x] ~~possibly need a generatevote**s** RPC in addition to the generatevote RPC to save round trip times when voting multiple tickets~~ performance is ok after tweaking generatevote implementation
- [x] audit locking

PR #155, opened 2017-04-27
put voting back into a single loop

jolan:
Part of #132.

### DIFF
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -5,6 +5,7 @@
 package main
 
 import (
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


backend/stakepoold/server.go
```diff
@@ -351,19 +347,83 @@ func (ctx *appContext) processWinningTickets(wt WinningTicketsForBlock) {
...
+			// Use defaults if not found.
+			log.Warnf("vote config not found for %v using default",
+				msa)
+			defaultVoteCfg := &userdata.UserVotingConfig{
+				Userid:          0,
+				MultiSigAddress: msa,
+				VoteBits:        ctx.votingConfig.VoteBits,
+				VoteBitsVersion: ctx.votingConfig.VoteVersion,
+			}
+			voteCfg = *defaultVoteCfg
```
chappjc: Why not?
```go
voteCfg = userdata.UserVotingConfig{...
```
But I was wondering if having to do this allocation (`&userdata.UserVotingConfig{`) on every loop is perhaps not great.  This never changes, so it could even be global const.

PR #163, opened 2017-04-29
Fix processWinningTickets benchmark

chappjc:
Benchmark was broken because appContext.votingConfig was nil.  So fix is to set a valid VotingConfig when creating the appContext in init() in server_test.go.

For a more realistic benchmark, also fill out the userVotingConfig map.  Previously it was empty, and the lookup would fail, and default voting config was used.  By adding random user multisigs, where the last 5 own the winning tickets.

Reorder fields of VotingConfig so it goes from 40 -> 32 bytes by avoiding excess padding.

Remove unnecessary allocation of a UserVotingConfig inside processWinningTickets (when msa lookup fails).

Simplify timing of processWinningTickets by using time.Since.

### DIFF
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -54,9 +54,9 @@ type appContext struct {
 // VotingConfig contains global voting defaults.
 type VotingConfig struct {
 	VoteBits         uint16
+	VoteVersion      uint32
 	VoteBitsExtended string
 	VoteInfo         *dcrjson.GetVoteInfoResult
-	VoteVersion      uint32
 }
 
 type WinningTicketsForBlock struct {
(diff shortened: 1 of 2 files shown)


backend/stakepoold/server.go
```diff
@@ -54,9 +54,9 @@ type appContext struct {
 // VotingConfig contains global voting defaults.
 type VotingConfig struct {
 	VoteBits         uint16
+	VoteVersion      uint32
 	VoteBitsExtended string
 	VoteInfo         *dcrjson.GetVoteInfoResult
-	VoteVersion      uint32
```
jolan: Can you remove VoteInfo? It should be unused now.

PR #164, opened 2017-05-01
stakepoold: make additional wallet calls async

marcopeereboom:
There is quite of bit of chatter between stakepoold and wallet.  In order to minimize network and wallet latency convert critical path wallet calls to async mode.

Ultimately we need the wallet to either batch these or do more independent work.  There should be not that amount of network chatter, period.  This is a lesson learned from SMB.


marcopeereboom:
Superseded by #165

PR #170, opened 2017-05-06
allow stakepoold on mainnet

jolan:
Pools are starting to migrate over so we can remove this restriction.


marcopeereboom:
@chappjc you did test this on mainnet?

jolan:
Agreed. We solicited feedback and help vetting the code on testnet.  We only received one set of results so not much more we can do but sort out issues at they come in.

PR #174, opened 2017-05-11
Add option to run stakepoold without gRPC server

chappjc:
This certainly would not be a recommended mode of operation, but with the current model of stakepoold connecting directly to MySQL for user voting configuration, it's possible to run an extra pool wallet without changing dcrstakepool config, while honoring user voting prefs.

The idea here is to reload every N minutes from MySQL if the `--norpclisten` flag is set.  This will also prevent the gRPC server from starting.

The first commit splits the timing of `vote()` into sign, send, and total.  This is unrelated to the gRPC change.

I'm pretty sure merging this is not the right idea, but I wanted to put it out there because it provides a quick way to fire up another wallet that will vote as desired without messing with the main server much, if at all.

Note there is also a doc correction in this (9109 -> 9113 and 19109 -> 19113).

### DIFF
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -312,11 +312,13 @@ func (ctx *appContext) loadData() {
 // winner contains all the bits and pieces required to vote and to print
 // statistics after usage.
 type winner struct {
-	msa      string                    // multisig
-	ticket   *chainhash.Hash           // ticket
-	config   userdata.UserVotingConfig // voting config
-	duration time.Duration             // overall vote duration
-	err      error                     // log errors along the way
+	msa          string                    // multisig
+	ticket       *chainhash.Hash           // ticket
+	config       userdata.UserVotingConfig // voting config
+	signDuration time.Duration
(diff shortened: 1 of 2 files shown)


backend/stakepoold/server.go
```diff
@@ -454,8 +473,9 @@ func (ctx *appContext) processWinningTickets(wt WinningTicketsForBlock) {
 			} else {
 				loserCount++
 			}
-			log.Infof("winning ticket %v msa %v duration %v: %v",
-				w.ticket, w.msa, w.duration, w.err)
+			log.Infof("winning ticket %v msa %v duration %v (%v + %v [+ %v]): %v",
+				w.ticket, w.msa, w.duration, w.signDuration, w.sendDuration,
+				w.duration-w.signDuration-w.sendDuration, w.err)
```
jolan: Will probably want to remove w.duration-w.signDuration-w.sendDuration in the future since so far it just looks like  the difference is precision/rounding.

PR #178, opened 2017-05-18
stakepoold: Fix ticket handling to be more accurate.

### LINKED ISSUE #177: stakepoold has extra tickets due to use of stakepooluserinfo
stakepoold is using stakepooluserinfo to retrieve ticket data.  This isn't ideal because:

1) It contains immature tickets in addition to live.
2) A few tickets have the wrong status (probably caused by missed notifications).
3) It requires being called for each user after every block is voted.

### LINKED ISSUE #179: stakepoold needs an on-disk cache/backup for tickets/user voting config
This would allow stakepoold to continue to work in a disaster case where MySQL is down/deleted/hacked/whatever.

I was thinking along the lines of dcrd's peers.json where we just save a backup to disk every 10 minutes or so.

Kind of related to #134. We still want to work regardless of what data source is down -- whether it be MySQL or dcrstakepool's gRPC client sending us updates.

jolan:
This reworks the ticket handling of stakepoold.

Previously, it would do stakepooluserinfo for every user on startup and every block thereafter.

This switches it to use gettickets on startup and then add/remove tickets based on the OnNewTickets / OnSpentAndMissedTickets notifications.

stakepoold's ticket count now stays in sync with the live ticket count from getstakeinfo.

This also adds caching so voting can resume immediately.

Closes #177.
Closes #179.

### DIFF
--- a/backend/stakepoold/ntfnhandlers.go
+++ b/backend/stakepoold/ntfnhandlers.go
@@ -7,15 +7,62 @@ import (
 	"github.com/decred/dcrrpcclient"
 )
 
+// checkIfBlockSeen increments the NeedToVote waitgroup by 1 if the block
+// has not been seen yet and records that it has been seen.
+func checkIfBlockSeen(ctx *appContext, ntfnName string, blockHash *chainhash.Hash, blockHeight int64) {
+	hasBeenSeen := true
+
+	ctx.Lock()
+	if !ctx.lastBlockSeenHash.IsEqual(blockHash) || ctx.lastBlockSeenHeight != blockHeight {
+		hasBeenSeen = false
+		ctx.wgNeedToVote.Add(1)
+		ctx.lastBlockSeenHash = blockHash
+		ctx.lastBlockSeenHeight = blockHeight
+	}
+	ctx.Unlock()
+
+	// Log notification information outside of the handler.
+	go func() {
+		log.Debugf("ntfn %s blockHeight %v blockHash %v", ntfnName, blockHeight,
+			blockHash)
+		if !hasBeenSeen {
+			log.Debugf("incremented wgNeedToVote for block height %v hash %v",
+				blockHeight, blockHash)
+		}
+	}()
+}
+
 // Define notification handlers
 func getNodeNtfnHandlers(ctx *appContext, connCfg *dcrrpcclient.ConnConfig) *dcrrpcclient.NotificationHandlers {
 	return &dcrrpcclient.NotificationHandlers{
-		OnWinningTickets: func(blockHash *chainhash.Hash, blockHeight int64,
-			winningTickets []*chainhash.Hash) {
+		OnNewTickets: func(blockHash *chainhash.Hash, blockHeight int64, stakeDifficulty int64, tickets []*chainhash.Hash) {
+			checkIfBlockSeen(ctx, "OnNewTickets", blockHash, blockHeight)
+			nt := NewTicketsForBlock{
+				blockHash:   blockHash,
+				blockHeight: blockHeight,
+				newTickets:  tickets,
+			}
+			ctx.newTicketsChan <- nt
+		},
+		OnSpentAndMissedTickets: func(blockHash *chainhash.Hash, blockHeight int64, stakeDifficulty int64, tickets map[chainhash.Hash]bool) {
+			checkIfBlockSeen(ctx, "OnSpentAndMissedTickets", blockHash, blockHeight)
+			ticketsFixed := make(map[*chainhash.Hash]bool)
+			for ticketHash, spent := range tickets {
+				ticketHash := ticketHash
+				ticketsFixed[&ticketHash] = spent
+			}
+			smt := SpentMissedTicketsForBlock{
+				blockHash:   blockHash,
+				blockHeight: blockHeight,
+				smTickets:   ticketsFixed,
+			}
+			ctx.spentmissedTicketsChan <- smt
+		},
+		OnWinningTickets: func(blockHash *chainhash.Hash, blockHeight int64, winningTickets []*chainhash.Hash) {
+			checkIfBlockSeen(ctx, "OnWinningTickets", blockHash, blockHeight)
 			wt := WinningTicketsForBlock{
 				blockHash:      blockHash,
 				blockHeight:    blockHeight,
-				host:           connCfg.Host,
 				winningTickets: winningTickets,
 			}
 			ctx.winningTicketsChan <- wt
--- a/backend/stakepoold/rpcclient.go
+++ b/backend/stakepoold/rpcclient.go
@@ -1,13 +1,16 @@
 package main
 
 import (
+	"encoding/hex"
 	"fmt"
 	"io/ioutil"
+	"time"
 
 	"github.com/decred/dcrd/chaincfg/chainhash"
 	"github.com/decred/dcrrpcclient"
 	"github.com/decred/dcrstakepool/backend/stakepoold/userdata"
 	"github.com/decred/dcrutil"
+	"github.com/decred/dcrwallet/wallet/txrules"
 )
 
 var requiredChainServerAPI = semver{major: 3, minor: 1, patch: 0}
@@ -110,70 +113,202 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
 	return dcrwClient, walletVer, nil
 }
 
-func walletFetchUserTickets(ctx *appContext) (map[chainhash.Hash]string, string) {
+// nodeCheckTicketFee evaluates a stake pool ticket to see if it's
+// acceptable to the stake pool. The ticket must pay out to the stake
+// pool cold wallet, and must have a sufficient fee.
+// This function is a port of evaluateStakePoolTicket in dcrwallet which plucks
+// the in/out amounts from the RPC response rather than the wallet's record.
+// Stake pool fee tests are present in dcrwallet/wallet/txrules.
+func nodeCheckTicketFee(ctx *appContext, txhex string, blockhash string) (bool, error) {
+	// Create raw transaction.
+	var buf []byte
+	buf, err := hex.DecodeString(txhex)
+	if err != nil {
+		return false, fmt.Errorf("DecodeString failed: %v", err)
+	}
+	res, err := ctx.nodeConnection.DecodeRawTransaction(buf)
+	if err != nil {
+		return false, fmt.Errorf("DecodeRawTransaction failed: %v", err)
+	}
+
+	bhash, err := chainhash.NewHashFromStr(blockhash)
+	if err != nil {
+		return false, fmt.Errorf("NewHashFromStr failed for %v: %v", blockhash, err)
+	}
+	gbh, err := ctx.nodeConnection.GetBlockHeader(bhash)
+	if err != nil {
+		return false, fmt.Errorf("GetBlockHeader failed for %v: %v", bhash, err)
+	}
+
+	// Check the first commitment output (Vout[1])
+	// and ensure that the address found there exists
+	// in the list of approved addresses. Also ensure
+	// that the fee exists and is of the amount
+	// requested by the pool.
+
+	commitAmt := dcrutil.Amount(0)
+	feeAddr := ""
+	feeAddrValid := false
+	ticketValue, err := dcrutil.NewAmount(res.Vout[0].Value)
+	if err != nil {
+		return false, fmt.Errorf("NewAmount failed: %v", err)
+	}
+
+	for i := range res.Vout {
+		switch res.Vout[i].ScriptPubKey.Type {
+		case "sstxcommitment":
+			for j := range res.Vout[i].ScriptPubKey.Addresses {
+				feeAddr = res.Vout[i].ScriptPubKey.Addresses[j]
+				_, exists := ctx.feeAddrs[feeAddr]
+				if !exists {
+					continue
+				}
+				feeAddrValid = true
+				feeAmount, err := dcrutil.NewAmount(*res.Vout[i].ScriptPubKey.CommitAmt)
+				if err != nil {
+					return false, fmt.Errorf("NewAmount failed: %v", err)
+				}
+				commitAmt = feeAmount
+				break
+			}
+		}
+	}
+
+	in := dcrutil.Amount(0)
+	for i := range res.Vout {
+		switch res.Vout[i].ScriptPubKey.Type {
+		case "sstxcommitment":
+			amount, err := dcrutil.NewAmount(*res.Vout[i].ScriptPubKey.CommitAmt)
+			if err != nil {
+				return false, fmt.Errorf("NewAmount failed: %v", err)
+			}
+			in += amount
+		}
+	}
+
+	out := dcrutil.Amount(0)
+	for i := range res.Vout {
+		amount, err := dcrutil.NewAmount(res.Vout[i].Value)
+		if err != nil {
+			return false, fmt.Errorf("NewAmount failed: %v", err)
+		}
+		out += amount
+	}
+	fees := in - out
+
+	if !feeAddrValid {
+		log.Warnf("Unknown pool commitment address %s for ticket %v",
+			feeAddr, res.Txid)
+		return false, nil
+	}
+
+	// Calculate the fee required based on the current
+	// height and the required amount from the pool.
+	feeNeeded := txrules.StakePoolTicketFee(ticketValue, fees,
+		int32(gbh.Height), ctx.feePercent, ctx.params)
+	if commitAmt < feeNeeded {
+		log.Warnf("User %s submitted ticket %v which has less fees than "+
+			"are required to use this stake pool and is being skipped "+
+			"(required: %v, found %v)", feeAddr, res.Txid, feeNeeded, fees)
+
+		// Reject the entire transaction if it didn't
+		// pay the pool server fees.
+		return false, nil
+	}
+
+	log.Debugf("Accepted valid stake pool ticket %v committing %v in fees",
+		res.Txid, commitAmt)
+
+	return true, nil
+}
+
+func walletGetTickets(ctx *appContext) map[chainhash.Hash]string {
 	// This is suboptimal to copy and needs fixing.
-	users := make(map[string]userdata.UserVotingConfig)
+	userVotingConfig := make(map[string]userdata.UserVotingConfig)
 	ctx.RLock()
 	for k, v := range ctx.userVotingConfig {
-		users[k] = v
+		userVotingConfig[k] = v
 	}
 	ctx.RUnlock()
 
+	ticketsIgnoredCount := 0
+	ticketsManuallyAddedCount := 0
+
 	userTickets := make(map[chainhash.Hash]string)
 
+	log.Info("Running GetTickets, this may take awhile...")
+	timenow := time.Now()
+	tickets, err := ctx.walletConnection.GetTickets(false)
+	log.Infof("GetTickets: took %v", time.Since(timenow))
+
+	if err != nil {
+		log.Warnf("GetTickets failed: %v", err)
+		return userTickets
+	}
+
 	type promise struct {
-		dcrrpcclient.FutureStakePoolUserInfoResult
-		msa string
+		dcrrpcclient.FutureGetTransactionResult
 	}
-	promises := make([]promise, 0, len(users))
-	for msa, v := range users {
-		addr, err := dcrutil.DecodeAddress(msa)
-		if err != nil {
-			log.Infof("Could not decode multisig address %v for %v: %v",
-				msa, v.Userid, err)
-			continue
-		}
+	promises := make([]promise, 0, len(tickets))
 
-		promises = append(promises, promise{
-			ctx.walletConnection.StakePoolUserInfoAsync(addr), msa})
+	log.Debugf("setting up GetTransactionAsync for %v tickets", len(tickets))
+	for _, ticket := range tickets {
+		// lookup ownership of each ticket
+		promises = append(promises, promise{ctx.walletConnection.GetTransactionAsync(ticket)})
 	}
 
-	var (
-		ticketcount, usercount int
-	)
-
+	counter := 0
 	for _, p := range promises {
-		spui, err := p.Receive()
+		counter++
+		log.Debugf("Receiving GetTransaction result for ticket %v/%v", counter, len(tickets))
+		gt, err := p.Receive()
 		if err != nil {
-			log.Errorf("unable to fetch tickets for user %v multisigaddr %v: %v",
-				users[p.msa].Userid, p.msa, err)
+			// All tickets should exist and be able to be looked up
+			log.Warnf("GetTransaction error: %v", err)
 			continue
 		}
-
-		if spui == nil || len(spui.Tickets) <= 0 {
-			continue
-		}
-
-		for _, ticket := range spui.Tickets {
-			if ticket.Status != "live" {
+		for i := range gt.Details {
+			_, ok := userVotingConfig[gt.Details[i].Address]
+			if !ok {
+				log.Warnf("Could not map ticket %v to a user, user %v doesn't exist", gt.TxID, gt.Details[i].Address)
 				continue
 			}
 
-			hash, err := chainhash.NewHashFromStr(ticket.Ticket)
+			hash, err := chainhash.NewHashFromStr(gt.TxID)
 			if err != nil {
-				log.Infof("invalid ticket %v", err)
+				log.Warnf("invalid ticket %v", err)
 				continue
 			}
 
-			userTickets[*hash] = users[p.msa].MultiSigAddress
-			ticketcount++
+			// stake pool mode in dcrwallet works by checking the fee when the
+			// ticket is first seen.  If the fee is too low, the ticket is
+			// placed in the invalid bucket.  Otherwise, the ticket is added
+			// to the wallet.
+			// When stake pool-specific functionality in dcrwallet is removed,
+			// most likely all tickets, whether they pay the correct fee or not,
+			// will end up in live tickets list.
+			// So we need to check if the ticket is in the low fee list first
+			// before doing normal fee processing.
+			_, isManuallyAdded := ctx.manuallyAddedTicketsMSA[*hash]
+			if isManuallyAdded {
+				ticketsManuallyAddedCount++
+				userTickets[*hash] = userVotingConfig[gt.Details[i].Address].MultiSigAddress
+			} else {
+				ticketFeesValid, err := nodeCheckTicketFee(ctx, gt.Hex, gt.BlockHash)
+				if ticketFeesValid {
+					userTickets[*hash] = userVotingConfig[gt.Details[i].Address].MultiSigAddress
+				} else {
+					ticketsIgnoredCount++
+					log.Warnf("ignoring ticket %v for msa %v ticketFeesValid %v err %v",
+						*hash, ctx.userVotingConfig[gt.Details[i].Address].MultiSigAddress, ticketFeesValid, err)
+				}
+			}
+			break
 		}
-		usercount++
 	}
 
-	ticketNoun := pickNoun(ticketcount, "ticket", "tickets")
-	userNoun := pickNoun(usercount, "user", "users")
+	log.Infof("tickets loaded -- total %v ignored %v manuallyadded %v",
+		len(userTickets), ticketsIgnoredCount, ticketsManuallyAddedCount)
 
-	return userTickets, fmt.Sprintf("loaded %v %v for %v %v",
-		ticketcount, ticketNoun, usercount, userNoun)
+	return userTickets
 }
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -5,13 +5,17 @@
 package main
 
 import (
+	"encoding/gob"
 	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/signal"
 	"path/filepath"
+	"reflect"
+	"runtime"
+	"strings"
 	"sync"
 	"time"
 
@@ -23,6 +27,7 @@ import (
 	"github.com/decred/dcrstakepool/backend/stakepoold/userdata"
 	"github.com/decred/dcrutil"
 	"github.com/decred/dcrutil/hdkeychain"
+	"github.com/decred/dcrwallet/wallet/txrules"
 	"github.com/decred/dcrwallet/wallet/udb"
 
 	_ "github.com/go-sql-driver/mysql"
@@ -30,26 +35,45 @@ import (
 
 type appContext struct {
 	sync.RWMutex
+
 	// locking required
-	ticketsMSA       map[chainhash.Hash]string            // [ticket]multisigaddr
-	userVotingConfig map[string]userdata.UserVotingConfig // [multisigaddr]
+	liveTicketsMSA          map[chainhash.Hash]string            // [ticket]multisigaddr
+	manuallyAddedTicketsMSA map[chainhash.Hash]string            // [ticket]multisigaddr
+	userVotingConfig        map[string]userdata.UserVotingConfig // [multisigaddr]
 
 	// no locking required
-	blockheight        int64
-	coldwalletextpub   *hdkeychain.ExtendedKey
-	dataPath           string
-	feeAddrs           map[string]struct{}
-	nodeConnection     *dcrrpcclient.Client
-	params             *chaincfg.Params
-	wg                 sync.WaitGroup // wait group for go routine exits
-	quit               chan struct{}
-	reloadTickets      chan struct{}
-	reloadUserConfig   chan struct{}
-	userData           *userdata.UserData
-	votingConfig       *VotingConfig
-	walletConnection   *dcrrpcclient.Client
-	winningTicketsChan chan WinningTicketsForBlock
-	testing            bool // enabled only for testing
+	coldwalletextpub       *hdkeychain.ExtendedKey
+	dataPath               string
+	feeAddrs               map[string]struct{}
+	feePercent             float64
+	newTicketsChan         chan NewTicketsForBlock
+	nodeConnection         *dcrrpcclient.Client
+	params                 *chaincfg.Params
+	lastBlockSeenHash      *chainhash.Hash
+	lastBlockSeenHeight    int64
+	wg                     sync.WaitGroup // wait group for go routine exits
+	wgNeedToVote           sync.WaitGroup
+	quit                   chan struct{}
+	reloadTickets          chan struct{}
+	reloadUserConfig       chan struct{}
+	spentmissedTicketsChan chan SpentMissedTicketsForBlock
+	userData               *userdata.UserData
+	votingConfig           *VotingConfig
+	walletConnection       *dcrrpcclient.Client
+	winningTicketsChan     chan WinningTicketsForBlock
+	testing                bool // enabled only for testing
+}
+
+type NewTicketsForBlock struct {
+	blockHash   *chainhash.Hash
+	blockHeight int64
+	newTickets  []*chainhash.Hash
+}
+
+type SpentMissedTicketsForBlock struct {
+	blockHash   *chainhash.Hash
+	blockHeight int64
+	smTickets   map[*chainhash.Hash]bool
 }
 
 // VotingConfig contains global voting defaults.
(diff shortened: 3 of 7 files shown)


backend/stakepoold/server.go
```diff
@@ -323,20 +355,48 @@ func (ctx *appContext) loadData() {
...
+	res, nt.err = ctx.walletConnection.GetTransaction(nt.ticket)
+	nt.getDuration = time.Since(start)
+	if nt.err != nil {
+		return
+	}
+	for i := range res.Details {
+		_, ok := ctx.userVotingConfig[res.Details[i].Address]
+		if ok {
+			// multisigaddress will match if it belongs a pool user
+			nt.msa = res.Details[i].Address
```
chappjc: Then break?

backend/stakepoold/rpcclient.go
```diff
@@ -111,66 +109,68 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
-			hash, err := chainhash.NewHashFromStr(ticket.Ticket)
+			hash, err := chainhash.NewHashFromStr(gt.TxID)
 			if err != nil {
 				log.Infof("invalid ticket %v", err)
 				continue
 			}
-
-			userTickets[*hash] = users[p.msa].MultiSigAddress
-			ticketcount++
+			userTickets[*hash] = ctx.userVotingConfig[gt.Details[i].Address].MultiSigAddress
```
chappjc: And break?
Also, I've seen a number of messed to ticket purchases where the outputs of the sstx have all the wrong indexes (e.g. swapped the Dc and Ds addresses).  I guess that's the idea here?  In case the op has done addticket for a non standard pool ticket?

backend/stakepoold/ntfnhandlers.go
```diff
@@ -10,12 +10,33 @@ import (
...
+		OnNewTickets: func(blockHash *chainhash.Hash, blockHeight int64, stakeDifficulty int64, tickets []*chainhash.Hash) {
+			nt := NewTicketsForBlock{
+				blockHash:   blockHash,
+				blockHeight: blockHeight,
+				newTickets:  tickets,
+			}
+			ctx.newTicketsChan <- nt
+		},
+		OnSpentAndMissedTickets: func(blockHash *chainhash.Hash, blockHeight int64, stakeDifficulty int64, tickets map[chainhash.Hash]bool) {
+			ticketsFixed := make(map[*chainhash.Hash]bool)
```
chappjc: Hashing the pointer seems like trouble.  How come this is needed?

backend/stakepoold/server.go
jolan: Yep, good catch.

backend/stakepoold/rpcclient.go
jolan: Yeah that was the idea.  I wasn't sure about the user error stuff.

backend/stakepoold/ntfnhandlers.go
jolan: I did that so ticketMetadata could be used for all the notification types.

backend/stakepoold/rpcclient.go
chappjc: I have to digest this a bit more, but I'm also thinking along the lines of not accidentally voting on malformed tickets (e.g. attempted cheaters) in addition to being sure to vote on the ones that were added manually.

jolan: Oh right, I guess this means I need to add fee enforcement checks to stakepoold now.  Let me know if you have any other gotchas to consider.

backend/stakepoold/ntfnhandlers.go
chappjc: By no means am I certain there's an issue, but two things feel funny. First is that memory leaks become much more likely with pointers, especially in the case of a map where you have buckets that need to be freed an so forth.  Second, how does a look up work?  Say you come in with another tx hash and want to see if it's in the table, can you get to it?  Hasn't the map hashed the pointer as a key, not the actual `chainhash.Hash`?

backend/stakepoold/server.go
```diff
@@ -367,36 +427,158 @@ func (ctx *appContext) vote(wg *sync.WaitGroup, blockHash *chainhash.Hash, block
...
+
+	// We use pointer because it is the fastest accessor.
+	smtickets := make([]*ticketMetadata, 0, len(smt.smTickets))
+
+	var wg sync.WaitGroup // wait group for go routine exits
+
+	ctx.wgVoting.Wait() // wait for voting to finish
+
+	ctx.RLock()
+	for ticket, spent := range smt.smTickets {
```
chappjc: The smTickets field still bugs me, being a `map[*chainhash.Hash]bool`.   It works as long as the only thing the map is used for is ranging over it to get the key-value pairs, but I'm afraid it might be used to lookup a hash, which won't work.

backend/stakepoold/server.go
```diff
@@ -367,36 +427,158 @@ func (ctx *appContext) vote(wg *sync.WaitGroup, blockHash *chainhash.Hash, block
...
+		sm := &ticketMetadata{
+			spent:  spent,
+			ticket: ticket,
```
chappjc: `ticketMetadata` could either have a non-pointer `chainhash.Hash` member or it could be referenced here.

backend/stakepoold/rpcclient.go
```diff
@@ -111,66 +109,68 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
-	}
-	ctx.RUnlock()
+	var (
+		ticketcount, usercount int
+	)
 
 	userTickets := make(map[chainhash.Hash]string)
 
+	log.Infof("Running GetTickets, this may take awhile...")
+	tickets, err := ctx.walletConnection.GetTickets(false)
```
chappjc: This takes a long time with 1000s of tickets.  I'm not sure this is practical unless `GetTickets` is accelerated.

backend/stakepoold/rpcclient.go
```diff
@@ -111,66 +109,68 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
 		}
-		usercount++
 	}
 
+	ctx.RUnlock()
+
+	usersWithTickets := make(map[string]int)
+
+	for _, address := range userTickets {
+		usersWithTickets[address] = 1
```
chappjc: Is this only for `usercount = len(usersWithTickets)` below or were there other plans?

backend/stakepoold/rpcclient.go
jolan: Yep I'm aware.  It should be able to be sped up a bit.

backend/stakepoold/rpcclient.go
jolan: Just for printing the user count.

jolan:
Well I added fee checking and it works OK in my testing so far.  I'm going to see what I can do to improve the speed of GetTickets and then clean this up some more.

chappjc:
@jolan Gonna test with your updates tonight if it's at a good spot.

jolan:
Yes it's ready for testing/review. CC @marcopeereboom

backend/stakepoold/server.go
```diff
@@ -301,46 +484,297 @@ func main() {
...
+
+	// getticket is used by processNewTickets and processSpentMissedTickets
+	// so we gracefully yield the processor until voting has completed
+	counter := 0
+	for {
+		lastVotedBlock := atomic.LoadInt64(&ctx.lastVotedBlock)
+		// XXX should really store and check the hash too but there's no
+		// atomic way to do that.  for now, fork blocks won't get vote
+		// prioritization.
+		if lastVotedBlock != nt.blockHeight {
```
chappjc: Might be safer to do `if lastVotedBlock < nt.blockHeight`

backend/stakepoold/server.go
```diff
@@ -371,43 +805,177 @@ func (ctx *appContext) vote(wg *sync.WaitGroup, blockHash *chainhash.Hash, block
...
+		ctx.liveTicketsMSA[ticket] = msa
+	}
+	ticketCountNew = len(ctx.liveTicketsMSA)
+	ctx.mtx.Unlock()
+	atomic.StoreUint64(&ctx.stats.TicketsLiveCount, uint64(ticketCountNew))
+
+	// Log ticket information outside of the handler.
+	go func() {
+		for ticket, msa := range addtickets {
+			atomic.AddUint64(&ctx.stats.TicketsNewCount, 1)
```
chappjc: Outside the loop: `atomic.AddUint64(&ctx.stats.TicketsNewCount, len(addtickets))`

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +112,170 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
+					if err != nil {
+						return false, fmt.Errorf("NewAmount failed: %v", err)
+					}
+					fees = feeAmount
+				}
+			}
+		}
+	}
+
+	in := dcrutil.Amount(0)
```
jolan: Done

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +112,170 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
+		amount, err := dcrutil.NewAmount(res.Vout[i].Value)
 		if err != nil {
-			log.Infof("Could not decode multisig address %v for %v: %v",
-				msa, v.Userid, err)
-			continue
+			return false, fmt.Errorf("NewAmount failed: %v", err)
 		}
+		out += amount
+	}
+	fees2 := in - out
```
jolan: I updated the comment to reflect that this function is ported from wallet where it does have some tests (but more for the underlying functions/calculations).  Needs some more though.

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +112,170 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
 
-			userTickets[*hash] = users[p.msa].MultiSigAddress
-			ticketcount++
+			// stake pool mode in dcrwallet works by checking the fee when the
+			// ticket is first seen.  If the fee is too low, the ticket is
+			// placed in the invalid bucket.  Otherwise, the ticket is added
+			// to the wallet.
+			// When stake pool-specific functionality in dcrwallet is removed,
+			// most likely all tickets, whether they pay the correct fee or not,
+			// will end up in live tickets list.
```
jolan: This won't happen in dcrwallet for awhile.  I am just planning for when that day comes and trying to explain the logic.

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +112,170 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
+				userTickets[*hash] = ctx.userVotingConfig[gt.Details[i].Address].MultiSigAddress
+			} else {
+				ticketFeesValid, err := nodeCheckTicketFee(ctx, gt.Hex, gt.BlockHash)
+				if ticketFeesValid {
+					userTickets[*hash] = ctx.userVotingConfig[gt.Details[i].Address].MultiSigAddress
+				} else {
+					atomic.AddUint64(&ctx.stats.TicketsIgnoredCount, 1)
+					log.Warnf("ignoring ticket %v for msa %v ticketFeesValid %v err %v",
+						*hash, ctx.userVotingConfig[gt.Details[i].Address].MultiSigAddress, ticketFeesValid, err)
+				}
```
jolan: Right it was just called from main. I ended up putting it back to a read lock since I may want to call gettickets more than just at startup at some point.

backend/stakepoold/server.go
```diff
@@ -242,18 +324,118 @@ func runMain() int {
...
+
+	ctx.printStats()
+
+	deferGetTickets := false
+
+	// get tickets now if we don't have any
+	if len(ctx.liveTicketsMSA) == 0 {
+		timenow := time.Now()
+		ctx.mtx.Lock()
+		ctx.liveTicketsMSA = walletGetTickets(ctx)
```
jolan: Right, if we're stuck doing GetTickets() because we have 0 tickets and a block comes in, we will miss it.

This is why I added caching -- so stakepoold can start immediately voting the tickets it loads from disk.  Then when gettickets finishes processing, it will delete/append the differences.

backend/stakepoold/server.go
```diff
@@ -301,46 +484,297 @@ func main() {
...
+			err = dec.Decode(&ctx.userVotingConfig)
+			if err != nil {
+				return err
+			}
+		}
+		log.Infof("Loaded %s from %s", dataKind, fullPath)
+		return nil
+	}
+
+	return nil
```
jolan: I did make this an error and made lastseen not an error.

backend/stakepoold/server.go
```diff
@@ -301,46 +484,297 @@ func main() {
...
+				log.Errorf("Failed to encode file %s: %v", ctx.dataPath, err)
+				continue
+			}
+		case "UserVotingConfig":
+			enc := gob.NewEncoder(w)
+			if err := enc.Encode(&ctx.userVotingConfig); err != nil {
+				log.Errorf("Failed to encode file %s: %v", ctx.dataPath, err)
+				continue
+			}
+		}
```
jolan: Done

backend/stakepoold/server.go
```diff
@@ -301,46 +484,297 @@ func main() {
...
-// winner contains all the bits and pieces required to vote and to print
-// statistics after usage.
-type winner struct {
+// ticketMetadata contains all the bits and pieces required to vote new tickets,
+// to look up new/missed/spent tickets, and to print statistics after usage.
+type ticketMetadata struct {
+	blockHeight  int64
 	msa          string                    // multisig
 	ticket       *chainhash.Hash           // ticket
+	spent        bool                      // spent (true) or missed (false)
```
jolan: Yeah I just copied how it was named in the notification. I need to look at how/where that is set a bit more closely I guess.

backend/stakepoold/server.go
jolan: Done

backend/stakepoold/server.go
jolan: Done (and found a couple more spots)

jolan:
I think functionality-wise this PR is good to go.

I updated my testnet stakepool setup and hit https://github.com/decred/dcrwallet/issues/825.

I'm also seeing an occasional hang when issuing gettickets().

It seems to happen 100% of the time with debuglevel=RPCC=debug.  When it hangs, I see the rpcclient output with the following output on the console:

```
2017-06-27 15:13:20.550 [INF] STPK: subscribed to notifications from dcrd
2017-06-27 15:13:20.550 [INF] STPK: Running GetTickets, this may take awhile...
2017-06-27 15:13:20.550 [INF] STPK: kicking off gettickets at 2017-06-27 15:13:20.550991638 -0500 CDT
2017-06-27 15:13:20.551 [TRC] RPCC: Sending command [gettickets] with id 3
2017-06-27 15:13:27.187 [TRC] RPCC: Received response for id 3 (result {"hashes":["1833c473a39c7d97fadb078151ccd4c9598dc009b87babe4bf19c30f40630c00",...
```

and in the log file:
```
2017-06-27 15:13:20.550 [INF] STPK: subscribed to notifications from dcrd
2017-06-27 15:13:20.550 [INF] STPK: Running GetTickets, this may take awhile...
2017-06-27 15:13:20.550 [INF] STPK: kicking off gettickets at 2017-06-27 15:13:20.550991638 -0500 CDT
2017-06-27 15:13:20.551 [TRC] RPCC: Sending command [gettickets] with id 3
```

So it looks like maybe the logger is hanging since the output doesn't match? Or maybe a locking issue?

backend/stakepoold/ntfnhandlers.go
```diff
@@ -10,12 +10,31 @@ import (
 // Define notification handlers
 func getNodeNtfnHandlers(ctx *appContext, connCfg *dcrrpcclient.ConnConfig) *dcrrpcclient.NotificationHandlers {
 	return &dcrrpcclient.NotificationHandlers{
-		OnWinningTickets: func(blockHash *chainhash.Hash, blockHeight int64,
-			winningTickets []*chainhash.Hash) {
+		OnNewTickets: func(blockHash *chainhash.Hash, blockHeight int64, stakeDifficulty int64, tickets []*chainhash.Hash) {
```
marcopeereboom: I would not be offended by some sort of wrapping here.

backend/stakepoold/rpcclient.go
```diff
@@ -1,13 +1,16 @@
 package main
 
 import (
+	"encoding/hex"
 	"fmt"
 	"io/ioutil"
+	"sync/atomic"
```
marcopeereboom: Danger, danger!  This almost certainly is abused.  When synchronizing sections you pretty much always want to use a mutex.  atomic should only be used if you are creating your own locking types.

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +113,193 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
+	if err != nil {
+		return false, fmt.Errorf("NewAmount failed: %v", err)
+	}
+
+	for i := range res.Vout {
+		switch res.Vout[i].ScriptPubKey.Type {
+		case "sstxcommitment":
+			for j := range res.Vout[i].ScriptPubKey.Addresses {
+				_, exists := ctx.feeAddrs[res.Vout[i].ScriptPubKey.Addresses[j]]
+				if exists {
```
marcopeereboom: use continue idiom in order to unindent some.

backend/stakepoold/userdata/userdata.go
```diff
@@ -16,7 +16,7 @@ type DBConfig struct {
 
 // UserData stores the current snapshot of the user voting config.
 type UserData struct {
-	sync.RWMutex
+	mtx              sync.RWMutex
```
marcopeereboom: Don't use mtx; make it nameless so that the data structure itself can be locked.  That is the idomatic way.

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +113,193 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
+
+	return true, nil
+}
+
+func walletGetTickets(ctx *appContext) map[chainhash.Hash]string {
 	// This is suboptimal to copy and needs fixing.
-	users := make(map[string]userdata.UserVotingConfig)
-	ctx.RLock()
+	userVotingConfig := make(map[string]userdata.UserVotingConfig)
+	ctx.mtx.RLock()
```
marcopeereboom: use ctx.Rlock() please.  Mutexes should be nameless.

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +113,193 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
+				}
+			}
+			break
 		}
-		usercount++
 	}
 
-	ticketNoun := pickNoun(ticketcount, "ticket", "tickets")
-	userNoun := pickNoun(usercount, "user", "users")
+	atomic.AddUint64(&ctx.stats.TicketsIgnoredCount, uint64(ticketsIgnoredCount))
```
marcopeereboom: I am worried about this construct.  Are you sure this is correct?  I'd use a mutex to lock the data structure.  This is almost assuredly racing with something.  The value will be probably correct but the fact that you have 2 atomic operations in a row tells me that this is probably broken.

jolan:
I can reproduce the hang with RPCC=trace on master so I will open a separate issue for that.

backend/stakepoold/server.go
```diff
@@ -23,33 +28,75 @@ import (
...
 type appContext struct {
-	sync.RWMutex
+	// The following variable must only be used atomically.
+	// It is used to defer processing of the OnNewTickets and
+	// OnSpentAndMissedTickets notifications until the
+	// OnWinningTickets ntfn has been processed and voting
+	// has taken place.
+	lastVotedBlock int64
+
+	mtx sync.RWMutex
```
marcopeereboom: Don't name mutexes.

backend/stakepoold/server.go
```diff
@@ -242,18 +324,119 @@ func runMain() int {
 	log.Infof("Connected to dcrd (JSON-RPC API v%s) on %v",
 		nodeVer.String(), curnet.String())
 
-	_, height, err := nodeConn.GetBestBlock()
+	curHash, curBlock, err := nodeConn.GetBestBlock()
 	if err != nil {
 		log.Errorf("unable to get bestblock from dcrd: %v", err)
 		return 8
```
marcopeereboom: It really is time to make this function return and error instead of an int.  Then on the caller you do

if err := runMain(); err != nil {
  fmt.Fprintf(os.Stderr, "%v\n", err)
}

The numbers are duplicated and mean nothing.  This also guaranteed has some errors that aren't printed making debugging painful.

backend/stakepoold/server.go
```diff
@@ -242,18 +324,119 @@ func runMain() int {
...
+			log.Warnf("unable to load user voting preferences from disk "+
+				"cache: %v", err)
+		} else {
+			log.Infof("Loaded UserVotingConfig for %d users from disk cache",
+				len(userVotingConfig))
+		}
+	}
+
+	statsBlockHeight := atomic.LoadInt64(&ctx.stats.BlockHeight)
+	atomic.StoreInt64(&ctx.lastVotedBlock, statsBlockHeight)
```
marcopeereboom: boo atomic!

backend/stakepoold/server.go
```diff
@@ -242,18 +324,119 @@ func runMain() int {
...
+			}
+		}
+		// append new live tickets
+		for ticket, msa := range liveTicketsTemp {
+			_, exists := ctx.liveTicketsMSA[ticket]
+			if !exists {
+				ctx.liveTicketsMSA[ticket] = msa
+			}
+		}
+		atomic.StoreUint64(&ctx.stats.TicketsLiveCount, uint64(len(ctx.liveTicketsMSA)))
```
marcopeereboom: atomic+mutex is probably wrong.

backend/stakepoold/server.go
```diff
@@ -301,46 +485,304 @@ func main() {
...
+	for dataKind, dataVersion := range saveFiles {
+		var filesToPrune []string
+
+		if fileExists(ctx.dataPath) {
+			files, err := ioutil.ReadDir(ctx.dataPath)
+			if err != nil {
+				return err
+			}
+
+			for i, file := range files {
```
marcopeereboom: Consider using continue.

backend/stakepoold/server.go
```diff
@@ -301,46 +485,304 @@ func main() {
...
+				log.Debugf("entry %d => %s", i, file.Name())
+				if strings.HasPrefix(file.Name(), strings.ToLower(dataKind)) &&
+					strings.Contains(file.Name(), dataVersion) &&
+					strings.HasSuffix(file.Name(), ".gob") {
+					filesToPrune = append(filesToPrune, filepath.Join(ctx.dataPath, file.Name()))
+				}
+			}
+
+			if len(filesToPrune) > saveFilesToKeep {
```
marcopeereboom: Consider using continue.

backend/stakepoold/server.go
```diff
@@ -301,46 +485,304 @@ func main() {
...
 
-// loadData loads the saved data from the saved file.  If empty, missing, or
-// malformed file, just don't load anything and start fresh
-func (ctx *appContext) loadData() {
-	ctx.Lock()
-	defer ctx.Unlock()
+// saveData saves some appContext fields to a file so they can be loaded back
+// into memory at next run.
+func saveData(ctx *appContext) {
+	ctx.mtx.Lock()
```
marcopeereboom: boo mtx!

backend/stakepoold/server.go
```diff
@@ -301,46 +485,304 @@ func main() {
...
+}
+
+// getticket pulls the transaction information for a ticket from dcrwallet. This is a go routine!
+func (ctx *appContext) getticket(wg *sync.WaitGroup, nt *ticketMetadata) {
+	waitStart := time.Now()
+
+	// getticket is used by processNewTickets and processSpentMissedTickets
+	// so we gracefully yield the processor until voting has completed
+	counter := 0
+	for {
```
marcopeereboom: This loop is very much not obvious and should be rewritten.

backend/stakepoold/server.go
```diff
@@ -371,43 +813,177 @@ func (ctx *appContext) vote(wg *sync.WaitGroup, blockHash *chainhash.Hash, block
...
+	ticketCountNew := 0
+	ticketCountOld := 0
+
+	ctx.mtx.Lock()
+	ticketCountOld = len(ctx.liveTicketsMSA)
+	for ticket, msa := range addtickets {
+		ctx.liveTicketsMSA[ticket] = msa
+	}
+	ticketCountNew = len(ctx.liveTicketsMSA)
+	ctx.mtx.Unlock()
```
marcopeereboom: Not loving atomic+mtx.

backend/stakepoold/server.go
```diff
@@ -371,43 +813,177 @@ func (ctx *appContext) vote(wg *sync.WaitGroup, blockHash *chainhash.Hash, block
...
+	ctx.mtx.Lock()
+	ticketCountOld = len(ctx.liveTicketsMSA)
+	for _, ticket := range missedtickets {
+		delete(ctx.liveTicketsMSA, *ticket)
+	}
+	for _, ticket := range spenttickets {
+		delete(ctx.liveTicketsMSA, *ticket)
+	}
+	ticketCountNew = len(ctx.liveTicketsMSA)
+	ctx.mtx.Unlock()
```
marcopeereboom: Not loving atomic+mtx.

backend/stakepoold/server.go
```diff
@@ -461,53 +1037,88 @@ func (ctx *appContext) processWinningTickets(wt WinningTicketsForBlock) {
...
-	for {
-		select {
-		case <-ctx.reloadTickets:
-			start := time.Now()
-			newTickets, msg := walletFetchUserTickets(ctx)
-			end := time.Now()
+func (ctx *appContext) printStats() {
+	ctx.updateStats()
+
+	// TODO would be nice to differentiate expired from missed
```
marcopeereboom: This should be mutexed and not be atomic.  Those 10 lines are racing.

backend/stakepoold/server.go
chappjc: To avoid the atomic stuff with `ctx.lastVotedBlock` and this loop, there could be a gate keeper function that returns a future (or just a channel) that signals when the vote goroutine has handled the latest block and it's cool for `processNewTickets` and `processSpentMissedTickets` to do their thing.

backend/stakepoold/rpcclient.go
chappjc: yeah, I think `Stats` could embed `sync.Mutex` and lock here and in [`printStats`](https://github.com/decred/dcrstakepool/pull/178/files/0a3952906c008fe90749ba4c58482258ffeb73de#diff-db56c53fc6985bfc454b56d97b3284b5R1082)

backend/stakepoold/server.go
chappjc: Ideally it would just read like `ctx.waitForVotes()` to keep this part clean.

backend/stakepoold/server.go
```diff
@@ -371,43 +813,177 @@ func (ctx *appContext) vote(wg *sync.WaitGroup, blockHash *chainhash.Hash, block
...
+	ticketCountOld := 0
+
+	ctx.mtx.Lock()
+	ticketCountOld = len(ctx.liveTicketsMSA)
+	for ticket, msa := range addtickets {
+		ctx.liveTicketsMSA[ticket] = msa
+	}
+	ticketCountNew = len(ctx.liveTicketsMSA)
+	ctx.mtx.Unlock()
+	atomic.StoreUint64(&ctx.stats.TicketsLiveCount, uint64(ticketCountNew))
```
chappjc: It's gonna be pretty difficult to keep the stats fields consistent with each other.  Either it doesn't matter, or they all need to be set at once.  In this function just adding a mutex to Stats and locking it might work, but since there are a few other spots where `ctx.stats.TicketsLiveCount` is potentially updated, I'm not sure what a solution to that might be.

backend/stakepoold/rpcclient.go
jolan: Done.

backend/stakepoold/rpcclient.go
jolan: Done.

backend/stakepoold/rpcclient.go
jolan: Done.

backend/stakepoold/rpcclient.go
jolan: I decided to remove all the stats stuff since it was pretty half-baked and I'm not sure it was very useful in the grand scheme of things.

backend/stakepoold/userdata/userdata.go
jolan: That one has to stay -- gob encoder that is used for the cache will choke otherwise.

backend/stakepoold/server.go
jolan: Done.

backend/stakepoold/server.go
jolan: Done.

backend/stakepoold/server.go
jolan: Done.

backend/stakepoold/server.go
jolan: Done.

backend/stakepoold/server.go
jolan: Done.

backend/stakepoold/server.go
jolan: I rewrote the loop but I'm still going to try to make it more like chappjc's suggestion.

backend/stakepoold/server.go
jolan: Done.

backend/stakepoold/server.go
jolan: Done.

backend/stakepoold/server.go
jolan: Yeah I just dropped everything about the stats.

jolan:
atomic usage has been completely removed and I fixed all of the other nits that I could.  Still working well on testnet.

jolan:
Hmm, using a channel works for awhile but eventually goes off the rails. I've reverted back for now.

chappjc:
I was thinking that a `WaitGroup` might make implementing this easier.  Just a thought, haven't worked anything out in detail.

jolan:
After switching to sending 1 time to 2 channels instead of sending 2 times to 1 channel, it appears to be working a lot better now.

jolan:
Just to recap a bit:

The wrinkle with doing the ticket accounting this way is that we now have 3 notifications to process instead of 1.  The notifications come in essentially at the same time but in somewhat random order.

With the atomic locking, everything worked fine and it prioritized votes well.

To get rid of the atomic locking, I tried channels but my implementation had bugs and wasn't prioritizing votes very well.

So I went to waitgroups which work well but didn't prioritize votes very well since typically processNewTickets or processSpentMissedTickets would fire before processWinningTickets and we'd fire up GetTransaction calls immediately before voting could even start spinning up.

So I decided just to sleep processNewTickets/processSpentMissedTickets for 100ms to give processWinningTickets a head start.

Now things are running smoothly with no atomic usage present.  It's been running stably for ~40 hours and the ticketcount matches getstakeinfo output exactly. And votes are still going out within ~10ms on my test setup.

backend/stakepoold/userdata/userdata.go
marcopeereboom: Why can it not be nameless?

backend/stakepoold/server.go
```diff
@@ -174,86 +219,156 @@ func runMain() int {
...
-	_, height, err := nodeConn.GetBestBlock()
+	// prune save data
+	err = pruneData(ctx)
 	if err != nil {
-		log.Errorf("unable to get bestblock from dcrd: %v", err)
-		return 8
+		log.Warnf("pruneData error: %v", err)
+	}
+
+	// not used currently
```
marcopeereboom: please clean up comments below.

backend/stakepoold/server.go
```diff
@@ -368,45 +728,178 @@ func (ctx *appContext) vote(wg *sync.WaitGroup, blockHash *chainhash.Hash, block
...
+		log.Infof("vote err %v", err)
+		w.err = err
+	} else {
+		w.txid = tx
+	}
 	w.sendDuration = time.Since(startSend)
 }
 
+func (ctx *appContext) processNewTickets(nt NewTicketsForBlock) {
+	time.Sleep(votingTargetTime) // give processWinningTickets time to vote
```
marcopeereboom: Sleep is not the solution; please use a signaling mechanism.  Sleep is the new goto.

backend/stakepoold/userdata/userdata.go
jrick: https://play.golang.org/p/tnqaa1udVN

backend/stakepoold/server.go
jolan: I made the mutex nameless.  As I said previously, it had problems with the gob encoder but I forgot I rewrote it a bit so it's no longer a problem.

I also cleaned up the comments.

backend/stakepoold/server.go
jolan: I was able to find a solution that didn't involve sleeping.

jolan:
Latest changes continued to run OK over the weekend.

chappjc:
I've not tested the current code, but the latest changes to the waitgroup usage feel like a more elegant than the previous solution.

chappjc:
Since this should be OK to test on a subset of stakepoold nodes, I'll give it a whirl on one mainnet wallet tonight and see how it does for 24 hrs or so.  I've seen many screwy sstx that the pool would vote on (mainly tickets manually added), but I'm a bit nervous about the validation code in these cases.  dcrwallet didn't care at voting time, only when deciding to add it to the stake pool manager, which of course is overridden by addticket.

jolan:
Squashed and rebased to cleanup a bit.

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +113,201 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
+		if err != nil {
+			return false, fmt.Errorf("NewAmount failed: %v", err)
+		}
+		out += amount
+	}
+	fees := in - out
+
+	if !feeAddrValid {
+		log.Warnf("Unknown pool commitment address %s for ticket %v",
+			feeAddr, res.Txid)
```
chappjc: `feeAddr` will not be set, will still be "" that is.  Also,

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +113,201 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
+				if !exists {
+					continue
+				}
+				feeAddr = res.Vout[i].ScriptPubKey.Addresses[j]
+				feeAddrValid = true
+				feeAmount, err := dcrutil.NewAmount(*res.Vout[i].ScriptPubKey.CommitAmt)
+				if err != nil {
+					return false, fmt.Errorf("NewAmount failed: %v", err)
+				}
+				commitAmt = feeAmount
```
chappjc: I think this should break out of the for loop here, otherwise this will get the last fee address for a sstxcommitment output type (not that someone would really have 2 fee addresses at this point).  As your comment says it should be vout 1

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +113,201 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
+		return false, fmt.Errorf("NewAmount failed: %v", err)
+	}
+
+	for i := range res.Vout {
+		switch res.Vout[i].ScriptPubKey.Type {
+		case "sstxcommitment":
+			for j := range res.Vout[i].ScriptPubKey.Addresses {
+				_, exists := ctx.feeAddrs[res.Vout[i].ScriptPubKey.Addresses[j]]
+				if !exists {
+					continue
```
chappjc: If we want to know what the bad address was a vout 1, we need to store it here before continuing

backend/stakepoold/rpcclient.go
```diff
@@ -110,70 +113,201 @@ func connectWalletRPC(cfg *config) (*dcrrpcclient.Client, semver, error) {
...
+			"are required to use this stake pool and is being skipped "+
+			"(required: %v, found %v)", feeAddr, res.Txid, feeNeeded, fees)
+
+		// Reject the entire transaction if it didn't
+		// pay the pool server fees.
+		return false, nil
+	}
+
+	log.Debugf("Accepted valid stake pool ticket %v committing %v in fees",
+		res.Txid, fees)
```
chappjc: This is printing the transaction/mining fee.  Is the intent to print the commitAmt?

backend/stakepoold/rpcclient.go
jolan: Stored now.

backend/stakepoold/rpcclient.go
jolan: Good point, done.

backend/stakepoold/rpcclient.go
jolan: Set now so it will be printed.

backend/stakepoold/rpcclient.go
jolan: Yep, fixed, thanks.

chappjc:
I have logs for about a week of running this with one wallet.  No errors, but I haven't compared with older versions of stakepoold logs to see if it's voting on every ticket.

jolan:
I've seen my overall missed rate drop from 1.6% to 1.0% on testnet since I started development on this branch.  Unfortunately, it was difficult to keep better metrics due to the time elapsed, number of changes made, and doing development on other branches based off of this one.

Hard to tell how much of that is getting (more accurate?) ticket data via gettickets instead of stakepooluserinfo, the better vote prioritization in this branch, or the new stake algorithm causing more ticket purchases which causes the mining templates to refresh more, etc.

Anyway, this needs to go in since we want to continue to improve the stake code in dcrwallet like https://github.com/decred/dcrwallet/pull/886.  Removing stakepooluserinfo from stakepoold serves as a first step for this.

PR #182, opened 2017-06-04
Network Stake Info API endpoint

chappjc:
The idea is to provide some pertinent information to API clients.

Needs:

- [ ] some docs.
- [ ] handle non-working testnet implementation of EstimateStakeDiff.


chappjc:
Totally forgot about this.  The idea here was to support a hypothetical stakepool app or something that could use this information in addition to the user's ticket data.

Issue #56, opened 2016-11-11
paginate/improve performance of tickets page


jolan:
The ticket page becomes slower to load over time because stakepooluserinfo returns every voted ticket.

Using my development/testnet stakepool, I can load the page OK for a user with 100,000 voted tickets with 3 wallets hosted on a LAN with very fast machines.

But once you have a few thousand on 3+ servers spread across the world on slowish VPSes you hit timeouts like @Dyrk reported seeing.

Real fix would be to either add sorting/pagination options to the RPC or overhaul the architecture of stakepool to cache the ticket info in the database.

Maybe the easiest thing to do would be to add an option to the stakepooluserinfo that only returns tickets of a certain status so voted tickets could be processed separately.

Dyrk:
I've fixed this one in my pool. I've totally removed all tickets related logic from the Tickets() route.
So, when user loading the page, he is seeing only Instructions.
Then I've implemented new API endpoint `stakepooluserinfo` where authorized user can retrieve his  tickets.
With Ajax request I get all the info and then display it to the user.

So, now most part of the logic is asynchronous and managed by frontend for this page. If you like this solution, I can make a pull request, however, it won't be easy due to many changes from the frontend and template side.

![img](https://dcrstats.com/img/uploads/spool-tickets-page.png)

jolan:
@Dyrk Yes we'd like that solution so if you can make a PR that'd be great.

chappjc:
even snippets of the AJAX parts would be helpful

Dyrk:
@chappjc here is the client side JS:
```
			/* Retreive Tickets from the API */
			$(document).ready(function() {
				var xsrf_token = "";
				var cookies = document.cookie.split(';');
				 for (var i = 0; i < cookies.length; i++) {
				   if (cookies[i].trim().match(/XSRF-TOKEN=(.*)$/g)) {
				 		xsrf_token = cookies[i].trim().replace('XSRF-TOKEN=', '');
				   }
				 }
				$.ajax({
					"url" : "/api/v0.1/stakepooluserinfo",
					"type" : "GET",
					beforeSend: function (request) {
						 request.setRequestHeader("X-XSRF-TOKEN", xsrf_token);
				  },
					success : function(data) {
						var json = {};
						try {
							json = JSON.parse(data);
						} catch(e) {
							showError(data); return;
						}

						if (json.status == "success") {
							displayTickets(json.data);
							$(".additional-tickets-info").show();
[...]

PR #183, opened 2017-06-04
Accelerate tickets page

chappjc:
Aims to address, in part, issue https://github.com/decred/dcrstakepool/issues/56

Reduced scope of this PR:

 - [x] Support returning a truncated limited list of the tickets, especially ancient voted tickets.  Doing this may be done either by sorting (which also gives a pre-sorted list to html/js) or by thresholding by spent age and only adding recent enough tickets to the html template data (the slices).  The latter is implemented in this PR, but the code to support ticket sorting is also added.
 - [x] The tickets page uses maps when not necessary and slices would be more efficient.  Slices also allow sorting (which may be commented out) for different ticket types.
 - [x] When using `StakePoolUserInfo` to get the tickets list for display on the /tickets page, do not bother getting it from all wallets and performing they "syncness" check.  This should be left to startup checks (or another maintenance routine), and should not hold up display of the web site when a valid response comes from the first (make it the fastest one!) wallet.

Possible goals for follow-up changes:

- Cache live tickets, but not voted. For each user.  Memory is not a constraint for millions of tickets, and CPU will not be either as the users will be in their own buckets.
- Put voted tickets on a separate page: (a) all voted on separate page and none on main tickets page, or (b)  all on the other page and with the most recent on main tickets page.
- To do (b) in the previous, cache the most recent voted tickets for each user.
- Allow downloading a CSV (or whatever) list of historical tickets.
- Maintaining the live/voted/etc tickets caches intelligently (i.e. without simply regenerating the lists if X minutes have passed) is desirable but not required.
- Write benchmarks for the sorters so it is clear when it becomes a burden on the server. (seems to be around 700 tickets in a slice).

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -74,6 +74,7 @@ type MainController struct {
 	version              string
 	voteVersion          uint32
 	votingXpub           *hdkeychain.ExtendedKey
+	maxVotedAge          int64
 }
 
 func randToken() string {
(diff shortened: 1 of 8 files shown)


chappjc:
rebased.

controllers/main.go
```diff
@@ -1692,46 +1729,70 @@ func (controller *MainController) Tickets(c web.C, r *http.Request) (string, int
...
 
-		for idx, ticket := range spui.InvalidTickets {
-			ticketInfoInvalid[idx] = TicketInfoInvalid{ticket}
+		for _, ticket := range spui.InvalidTickets {
+			ticketInfoInvalid = append(ticketInfoInvalid, TicketInfoInvalid{ticket})
 		}
 	}
 
+	// Sort live tickets
+	sort.Sort(ByTicketHeight(ticketInfoLive))
```
chappjc: Might as well comment this.  At best it will help accelerate the JS sorting by providing a pre-sorted list.

chappjc:
Darn you go 1.7.x
Yeah, I'll fix it up somehow

chappjc:
For the record:

```BenchmarkSortByTicketHeight100-32                 300000              5669 ns/op
BenchmarkSortByTicketHeight500-32                  30000             48334 ns/op
BenchmarkSortByTicketHeight1000-32                 10000            122387 ns/op
BenchmarkSortByTicketHeight2500-32                  3000            346371 ns/op
BenchmarkSortByTicketHeight5000-32                  2000            760506 ns/op
BenchmarkSortByTicketHeight10000-32                 1000           1714685 ns/op
BenchmarkSortByTicketHeight20000-32                  300           3771348 ns/op
BenchmarkSortBySpentByHeight100-32                200000              6407 ns/op
BenchmarkSortBySpentByHeight500-32                 30000             56892 ns/op
BenchmarkSortBySpentByHeight1000-32                10000            136414 ns/op
BenchmarkSortBySpentByHeight2500-32                 3000            396987 ns/op
BenchmarkSortBySpentByHeight5000-32                 2000            890614 ns/op
BenchmarkSortBySpentByHeight10000-32                 500           2055087 ns/op
BenchmarkSortBySpentByHeight20000-32                 300           4433991 ns/op```

PR #184, opened 2017-06-10
bump requried node and wallet RPC API versions

chappjc:
For Decred release 1.0.4

### DIFF
--- a/backend/stakepoold/rpcclient.go
+++ b/backend/stakepoold/rpcclient.go
@@ -10,8 +10,8 @@ import (
 	"github.com/decred/dcrutil"
 )
 
-var requiredChainServerAPI = semver{major: 3, minor: 0, patch: 0}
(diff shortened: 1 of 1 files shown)


chappjc:
I can't get to that now, but I understand the goal

chappjc:
@jolan Note that I added a debug logging statement for the `StakePoolUserInfo` call on /tickets:

https://github.com/decred/dcrstakepool/pull/183/files#diff-94feda3a91613ce92f105b5f66899dc8R1731

It's very slow.  Like 2-5 seconds for a few thousand tickets.  I know gettickets is sluggish too, but that's only used in the startup resync at the moment.  Although I recall there was discussion of fixing up gettickets to serve the purpose presently handled by stakepooluserinfo.  I feel like a very different api, like getvotedtickets/getlivetickets/etc or even an argument like gettickets(tickettype TicketTypeBitMask) might be needed

PR #189, opened 2017-06-20
Update deps to pull in new logger.

jrick:


jrick:
Fixed by adding a check for a non-nil rotator variable in the writer impl.

PR #196, opened 2017-07-14
mention wallet/script dependency and voting docs

### LINKED ISSUE #195: When importing api key in gui wallet, must use same wallet that you signed up with
Not sure if this is the best place for this issue (maybe docs) but it comes up enough that we do need to address it somewhat.

Multiple users have signed up for the pool, setup with one wallet, then made a new wallet and tried to add the same stakepool and hit an invalid api key error.  The pool does say that you can't change your public key, but users definitely seem to miss that.  Either it needs to be more prominent, or we need to explain what that really means better, or maybe we need to add it to the docs site.  I'm open to suggestions.

jolan:
Mention that the generated multisigscript depends on both wallets and tell them to talk to the pool admin/support staff if they need to change it.

Closes #195.

Some other minor stuff:

- Point to voting docs in case they want to self-vote with the CLI wallet.
- Update example price to be more realistic.
- No longer explicitly recommend auto-ticketbuying over manual now that the ticket pricing algorithm is more sane.

### DIFF
--- a/views/address.html
+++ b/views/address.html
@@ -2,14 +2,33 @@
 <div class="wrapper">
     <div class="row">
   <div class="col-xs-15 col-md-8 col-lg-8 notication-col center-block">
-    {{range .Flash}}<div class="well well-notification  orange-notification">{{.}}</div>{{end}}
+    {{range .Flash}}<div class="well well-notification orange-notification">{{.}}</div>{{end}}
   </div>
     <div class="col-sm-15 col-md-10 text-left center-block">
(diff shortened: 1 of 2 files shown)


chappjc:
Maybe I think differently than most users, but I feel like the new ticket price algorithm finally makes auto-ticketbuying make sense (you can buy right away when your tickets vote).  Previously, manual buying was much more sensible since there was a relatively small ticket buying window with hard-to-predict fees.

jolan:
The problem was people were manually buying tickets with low fees and no expiration during periods of high fees which is why we recommended ticket buyer over manual purchasing.

In any event, I don't see any reason to recommend one method over the other anymore.

PR #204, opened 2017-08-28
Test against go 1.9 and drop 1.7

dajohi:


run_tests.sh
```diff
@@ -0,0 +1,55 @@
...
+  --deadline=10m ./... | tee /dev/stderr)\"&& \
+  env GORACE='halt_on_error=1' go test -short -race \
+  \$(glide novendor)"
+
+if [ $GOVERSION == "local" ]; then
+    eval $TESTCMD
+    exit
+fi
+
+DOCKER_IMAGE_TAG=decred-golang-builder-$GOVERSION
```
chappjc: I'm not familiar with the new docker test env.  I suppose this image still needs to be made in https://github.com/decred/dcrdocker?

dajohi: sh run_tests.sh local

PR #212, opened 2017-09-15
remove voting prioritization

### LINKED ISSUE #211: stakepoold panics when wallet goes down and comes back
Maybe when a block goes by while wallet is down.

I don't have the stack trace at the moment, but it happened twice on the current version in the last couple days while I was starting and restarting things.  Fortunately it was a clear issue on [this line](https://github.com/decred/dcrstakepool/blob/master/backend/stakepoold/server.go#L954):

    ctx.wgNeedToVote.Done()

in `processWinningTickets`.  The panic was a negative waitGroup counter.

This seems to be because it always calls `Done()` for the wg even if `Add(1)` is not called because of the conditional [here](https://github.com/decred/dcrstakepool/blob/master/backend/stakepoold/ntfnhandlers.go#L16-L21):

```go
if !ctx.lastBlockSeenHash.IsEqual(blockHash) || ctx.lastBlockSeenHeight != blockHeight {
	hasBeenSeen = false
	ctx.wgNeedToVote.Add(1)
	ctx.lastBlockSeenHash = blockHash
	ctx.lastBlockSeenHeight = blockHeight
}
```

jolan:
We went back and forth in #178 on different ways to implement prioritize handling of the winning tickets notification/voting over the other notifications.

As we can see from #211, a waitgroup is not the right solution. It essentially makes us stateful but that doesn't work since dcrwallet may send us 0 or 2 voting notifications, rather than the 1 we expect to get 99.9% of the time.

~~Opened for discussion. This is a WIP (mostly because I haven't tested it yet)~~

Closes #211.

### DIFF
--- a/backend/stakepoold/ntfnhandlers.go
+++ b/backend/stakepoold/ntfnhandlers.go
@@ -7,36 +7,10 @@ import (
 	"github.com/decred/dcrrpcclient"
 )
 
-// checkIfBlockSeen increments the NeedToVote waitgroup by 1 if the block
-// has not been seen yet and records that it has been seen.
-func checkIfBlockSeen(ctx *appContext, ntfnName string, blockHash *chainhash.Hash, blockHeight int64) {
-	hasBeenSeen := true
-
-	ctx.Lock()
-	if !ctx.lastBlockSeenHash.IsEqual(blockHash) || ctx.lastBlockSeenHeight != blockHeight {
-		hasBeenSeen = false
-		ctx.wgNeedToVote.Add(1)
-		ctx.lastBlockSeenHash = blockHash
-		ctx.lastBlockSeenHeight = blockHeight
-	}
-	ctx.Unlock()
-
-	// Log notification information outside of the handler.
-	go func() {
-		log.Debugf("ntfn %s blockHeight %v blockHash %v", ntfnName, blockHeight,
-			blockHash)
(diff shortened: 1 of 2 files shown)


jolan:
I ran this over the weekend on my testnet voting wallets and didn't encounter any known issues.  Will continue to test and review the logs to make sure but it's looking good thus far.

jolan:
I don't think that the voting prioritization is still needed.  It was required on my original test setup (go 1.8 w/non-SSD to mimic the worst case scenario) to keep voting within 100ms.  Since migrating my test nodes to SSD and upgrading go to 1.9, I no longer see huge peaks in voting times. 

processWinningTickets is finishing in under 10ms on my hardware 90% of the time with and without sleeping to goroutines.  It held up during 2 periods of ticket buying where I was adding 10 tickets per block.  I think part of this is due to the continued improvements to the garbage collector in go 1.9 and obviously the other part is using a sensible storage medium for low latency voting.

I still do see a few peaks like usual in the 20-85ms range.  With 3 processes involved, I guess it depends on when the garbage collector for each process runs and the luck of how that works out.

So I'd like to just remove the voting prioritization for now.  If latency becomes an issue on production pools then @marcopeereboom and I have discussed some better solutions and it can be addressed in a separate PR.

chappjc:
sounds good, looks good

PR #215, opened 2017-09-22
glide up, bump version, release notes

jolan:
~~Still doing final testing but this should be it.~~ Testing complete.


jolan:
Good point, I'll add that to this PR since I want to start tagging releases.

chappjc:
Getting this, but it's not an issue:

    [WARN]	Lock file may be out of date. Hash check of YAML failed. You may need to run 'update'

jolan:
I probably mangled it when manually editing out the 3rd party repo changes.

Issue #201, opened 2017-07-31
Voted Ticket Notification


pedrotst:
[Feature Request]
It would be nice to have some sort of notification system to notify when a ticket was voted. This way we don't have to check the pool website everyday to check that. Specially now that we don't have buy windows anymore.

jolan:
A couple of the pools have this feature implemented already.  For the reference implementation, maybe another standalone program that could live in backend.

jolan:
A couple of the pools have this feature implemented already.  For the reference implementation, maybe another standalone program that could live in backend.

chappjc:
@jolan Yeah, that sounds like a good idea to keep it separate.  Just a little app that waits for voted notifications, looks up the email for the redeemed ticket multisig, and send an email.  It's the part where it looks up emails that I'm not so sure about since what we ultimately want to for MySQL to be untouched by anything but dcrstakepool

jolan:
This is needed so we can mark low fee tickets as voted in the database (continuing on #219) and also so we can remove any reference to stakepooluserinfo on the voting page / getstakeinfo on the stats page (#180).

xaur:
Please add 'privacy' label. Notifications have a privacy aspect: the service that you receive notification to (e.g. your email provider) will learn about your holdings and voting preferences. There should be a way to encrypt notifications to recipient's public key.

kiahmed:
have this ever been implemented or any script that could be hooked up?

Issue #180, opened 2017-05-22
stakepooluserinfo use is problematic when adding wallets


chappjc:
The number of redeem scripts (from `listscripts`) and tickets (from `gettickets`) may be the same on all wallets, but it is possible for `getstakeinfo` to not match for several reasons that are difficult to rectify.

One is that if the operator has performed `addtickets` on the wallets at some point, it is necessary to do add the same tickets (until they've expired) on newly added wallets to get the number of `"live"` tickets in `getstakeinfo` to match between wallets.

In the same case as above, even after doing `addtickets` (for every invalid ticket ever added), the  `"voted"` number on the new wallet does not catch up with the elder wallets.

Perhaps this should be a dcrwallet issue, but perhaps it's time to workaround this in dcrstakepool given the new stakepoold arch changes.

jolan:
The goal is to keep migrating functionality over to stakepoold.  I think stakepooluserinfo can eventually be removed.

Rough TODO would be:

1) Move fee enforcement to stakepoold (WIP #178)
2) Re-work tickets page.  Maybe go back to using ticketsforaddress and add DB caching.
3) Add some sort of admin/invalid tickets interface so those tickets can be stored in the database and transmitted to/pulled from stakepoold.
4) Remove stakepooluserinfo and the fee checking from dcrwallet.

jolan:
Item 3 is in progress too.

2 and 4 may have to be split out and planned for the next release cycle.  There are a lot of dcrwallet/stakepoold changes in-flight already.

jolan:
I edited the title to reflect the addition of getstakeinfo and to mention addticket. The problems really lie with the original implementation of the stake code in wallet.  That code has been improved and refactored and we need to cope with the changes.

Per https://github.com/decred/dcrwallet/pull/913:

Added tickets are auxiliary to transaction history and do not appear in getstakeinfo stats.

Current status:

- stakepooluserinfo is still in use by the tickets page.  Could possibly replace with the implementation of https://github.com/decred/dcrwallet/issues/833.
- getstakeinfo is still in use by the stats page.  Need to create a suitable replacement that accounts for tickets that were added via addticket.

jolan:
Not actively working on this at the moment but it still needs to be done.

PR #219, opened 2017-10-16
multi: improve handling of low fee tickets

### LINKED ISSUE #134: switch stakepoold to obtain user voting preferences via gRPC

### LINKED ISSUE #209: need a better way for stakepoold to pick up addticket'd tickets
Since stakepoold does an initial GetTickets() and then adds/subtracts tickets from block notifications, it has no way to pick up new tickets added via addticket other than restarting stakepoold.

Would be nice to provide a web interface for addticket that sends all the newly added ticket hashes to stakepoold or maybe just triggers a reload when done.

### LINKED ISSUE #218: improve low-fee/manually added tickets handling
Currently, it isn't clear in an exportable way which tickets have been added by the admin and which have not.

Need to do something like:

[ ] - add API/admin UI to dcrstakepool that allows adding/removing low-fee tickets to the DB
[ ] - prune table on dcrstakepool startup
[ ] - have stakepoold retrieve the list from the DB and cache it
[ ] - modify fee checks to consult the list

jolan:
This brings proper support for handling invalid/low fee tickets into the stakepool.

Basically it adds storage to the DB for saving added tickets, an admin UI to remove/add to the table, and then pushes/pulls the lists to/from stakepoold.

Not really sure about the gRPC server <-> main app ctx channel stuff.  Seems to work OK but maybe there's a better way to do it.

Closes #134.
Closes #209.
Closes #218.

Also makes the fee checking fast per discussions with @chappjc and @jrick.

This also paves the way for closing some of the other open issues like #180 & #201.

Works for me™ so far but needs more testing and review.


jrick:
Is this the last of the bits needed to remove the remaining stake pool code out of dcrwallet?

backend/stakepoold/server.go
```diff
@@ -178,6 +190,92 @@ func deriveChildAddresses(key *hdkeychain.ExtendedKey, startIndex, count uint32,
...
+	}
+
+	log.Debugf("Accepted valid stake pool ticket %v committing %v in fees",
+		tx.TxHash(), tx.TxOut[0].Value)
+
+	return true, nil
+}
+
+// MsgTxFromHex returns a wire.MsgTx struct built from the transaction hex string
+func MsgTxFromHex(txhex string) *wire.MsgTx {
```
chappjc: Probably best to return or log the error too.

jolan:
@jrick This lays the foundation for getting there. I think what's left on the stake pool side is:

- Improve the gRPC client code that dcrstakepool is using (there's a TODO there; still thinking what that should look like) so it is ready to be user facing.
- Switch user facing tickets page from **stakepooluserinfo** to using data from stakepoold gRPC.   This would provide "Live" and "Invalid" but not Missed/Voted/Expired.
  - Could remove the missed/voted/expired sections from the user-facing tickets page and tell the users to use decrediton's ticket history stuff instead (https://github.com/decred/decrediton/pull/745) or:
  - implement voted ticket notification (#201)

backend/stakepoold/rpc/rpcserver/server.go
```diff
@@ -17,16 +17,21 @@
...
+	"github.com/decred/dcrd/chaincfg/chainhash"
 	pb "github.com/decred/dcrstakepool/backend/stakepoold/rpc/stakepoolrpc"
+	"github.com/decred/dcrstakepool/backend/stakepoold/userdata"
 )
 
 // Public API version constants
 const (
-	semverString = "3.0.0"
-	semverMajor  = 3
+	grpcTimeout  = time.Millisecond * 100
```
marcopeereboom: Isn't that silly short?

backend/stakepoold/rpc/rpcserver/server.go
```diff
@@ -54,32 +59,256 @@ func (*versionServer) Version(ctx context.Context, req *pb.VersionRequest) (*pb.
...
+	case s.getAddedLowFeeTicketsChan <- struct{}{}:
+		log.Debug("GetAddedLowFeeTickets: signaled getAddedLowFeeTicketsChan")
+	default:
+		// We log this in order to detect if we potentially
+		// have a deadlock.
+		log.Debug("GetAddedLowFeeTickets: query not sent")
+		goto done
+	}
+
+	for {
```
jrick: this doesn't need to be a loop

backend/stakepoold/rpc/rpcserver/server.go
```diff
@@ -54,32 +59,256 @@ func (*versionServer) Version(ctx context.Context, req *pb.VersionRequest) (*pb.
...
+			goto done
+		case <-ctx.Done():
+			log.Debug("GetAddedLowFeeTickets: timed out %v", ctx.Err())
+			goto done
+		}
+	}
+
+done:
+	defer func() {
+		log.Infof("GetAddedLowFeeTickets processed %v tickets in %v waitTime %v",
```
jrick: not sure if this is common in this codebase but i think intercepting the requests (similar to how i do it in wallet) is a much cleaner and simpler way to keep track of this info

backend/stakepoold/rpc/rpcserver/server.go
jolan: No it's just copying of overwriting a map which is ~2ms on my machine.  I think it could be re-written to not need a timeout.

backend/stakepoold/rpc/rpcserver/server.go
jrick: https://github.com/decred/dcrwallet/blob/768542b7d2dd1d021c51cf3f07c2c24d46151cd3/rpcserver.go#L215

capture the current time before and after running the handler and log the delta

backend/stakepoold/rpc/rpcserver/server.go
```diff
@@ -54,32 +59,256 @@ func (*versionServer) Version(ctx context.Context, req *pb.VersionRequest) (*pb.
...
+	}()
+
+	return &pb.GetIgnoredLowFeeTicketsResponse{Tickets: tickets}, ctx.Err()
+}
+
+func (s *stakepooldServer) GetLiveTickets(ctx context.Context, req *pb.GetLiveTicketsRequest) (*pb.GetLiveTicketsResponse, error) {
+	startTime := time.Now()
+	tickets := make([]*pb.TicketEntry, 0)
+
+	// limit the time we take
```
jrick: Use context.WithDeadline

backend/stakepoold/rpc/api.proto
```diff
@@ -3,24 +3,61 @@ syntax = "proto3";
...
+message SetUserVotingPrefsResponse {
+	bool Processed = 1;
+}
+message SetUserVotingPrefsRequest {
+	repeated UserVotingConfigEntry user_voting_config = 1;
+}
+
+message TicketEntry {
+	string TicketAddress = 1;
+	string TicketHash = 2;
```
jrick: why a string instead of bytes?

backend/stakepoold/rpc/rpcserver/server.go
```diff
@@ -54,32 +59,256 @@ func (*versionServer) Version(ctx context.Context, req *pb.VersionRequest) (*pb.
...
+			"not sent")
+	}
+
+	defer func() {
+		log.Infof("StoreUserVotingPrefs: processed %v users in %v",
+			len(req.UserVotingConfig), time.Since(startTime))
 	}()
-	return &pb.UpdateVotingPrefsResponse{}, nil
+
+	return &pb.SetUserVotingPrefsResponse{Processed: processed}, nil
```
jrick: should return a non-nil error if it timed out and the request wasn't processed

backend/stakepoold/rpc/rpcserver/server.go
jrick: using an interceptor would also be a good place to add the deadline to the context

PR #221, opened 2017-10-20
Enforce minimum host count consistently

infertux:
The `MinServers` setting refers to the minimum number of wallets connected needed to avoid errors and should be enforced consistently for _wallet_ hosts and _stakepoold_ hosts. Moreover, this change makes it easier to run in development mode by setting `MinServers` to 1.

This change allows me to have a self-contained VM to run dcrstakepool in development mode by setting `minservers=1` in _.dcrstakepool/dcrstakepool.conf_.

### DIFF
--- a/config.go
+++ b/config.go
@@ -539,9 +539,9 @@ func loadConfig() (*config, []string, error) {
 	// Add default wallet port for the active network if there's no port specified
 	cfg.WalletHosts = normalizeAddresses(cfg.WalletHosts, activeNetParams.WalletRPCServerPort)
 
-	if len(cfg.WalletHosts) < 2 {
-		str := "%s: you must specify at least 2 wallethosts"
-		err := fmt.Errorf(str, funcName)
+	if len(cfg.WalletHosts) < cfg.MinServers {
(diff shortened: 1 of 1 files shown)


infertux:
> MinServers is the minimum number of wallets connected not configured so this isn't quite right.

Oops I've read too fast. My bad.

> And it isn't really possible to run a proper development environment without 2+ wallets right now due to some issues in the code.

I've been running this code for a while now and even though I get a large amount of missed tickets, I do get successful votes as well, which is good enough for development/testing purposes.

> To be honest, a lot of the wallet handling and ticket/script checking can be gutted or simplified to achieve what you want. I can write up an issue for it and help you implement it if you're interested.

Sure, I'm definitely interested. I think it's important to lower the barrier for contributing to dcrstakepool. Right now, it's pretty cumbersome to set up a dev environment if you need multiple voting wallets with different IPs, etc.

jolan:
There are now a bunch of open related issues surrounding improving this but the short version is that there really shouldn't be a wallet connection at all nor restrictions on anything but the number of voting hosts (and we could relax that to 1 on testnet).  Closing this in favor of future work.

PR #231, opened 2017-12-03
bugfix: init grpcConnections with 0 len

detailyang:
now the default `grpcConnections` is maked by `grpcConnections := make([]*grpc.ClientConn, len(cfg.StakepooldHosts))`, it may cause get nil dereference when stakepoold is disable by `enablestakepoold = false`

### DIFF
--- a/server.go
+++ b/server.go
@@ -105,7 +105,7 @@ func runMain() int {
 	// Supported API versions are advertised in the API stats result
 	APIVersionsSupported := []int{1, 2}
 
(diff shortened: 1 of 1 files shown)


dajohi:
@detailyang Where is the nil dereference?

detailyang:
@dajohi  when call the AdminStatus controller at [https://github.com/decred/dcrstakepool/blob/master/controllers/main.go#L971](https://github.com/decred/dcrstakepool/blob/master/controllers/main.go#L971)

detailyang:
@dajohi  This fix is completely wrong, I will send a new PR to fix it, sorry:(

PR #236, opened 2017-12-11
config: use cfg.HomeDir to replace defaultHomeDir

detailyang:
Now we use the defaultHomeDir to create dir, it could failed when user change the home dir to another.


detailyang:
@jolan  thanks, i will look at this tomorrow but now it's too late and i need go to sleep:)

detailyang:
@jolan Did i missing something. This fix is about stakepoold which always use default homeDir rather than AppData just like what you said

PR #239, opened 2017-12-24
remove tee from tests

dajohi:


jolan:
Closing in favor of #245.

PR #250, opened 2018-01-23
[WIP] stakepoold: fix newly matured tkt edge case

### LINKED ISSUE #249: tickets that mature and are called to vote in the same block are not voted
stakepoold acts on notifications from dcrd but since the winning tickets notification typically arrives before the new tickets notification, we won't properly vote when a newly matured ticket is called to vote since live tickets are only appended when they mature.

jolan:
As mentioned in the linked issue, there's an edge case where a ticket
can mature and be picked to vote in the same block.  When this happens
it's basically just the luck of the draw as to if the new ticket
notification arrives before the winning ticket notification.  If it
doesn't, then we will fail to vote since we aren't aware of its presence
when we go to vote it.  This keeps a map of unknown winning tickets and
will attempt to vote such tickets from the new ticket handler.

WIP, needs tests added and live testing (and probably some more refactoring).
This shouldn't be a frequent occurrence on mainnet but still needs to be
addressed since it has been hit in production.

Resolves #249.

### DIFF
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -44,6 +44,7 @@ type appContext struct {
 	addedLowFeeTicketsMSA   map[chainhash.Hash]string            // [ticket]multisigaddr
 	ignoredLowFeeTicketsMSA map[chainhash.Hash]string            // [ticket]multisigaddr
 	liveTicketsMSA          map[chainhash.Hash]string            // [ticket]multisigaddr
+	possiblyMissingTickets  map[chainhash.Hash]int64             // ticket]blockheight
 	userVotingConfig        map[string]userdata.UserVotingConfig // [multisigaddr]
 
 	// no locking required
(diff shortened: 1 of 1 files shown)


backend/stakepoold/server.go
```diff
@@ -44,6 +44,7 @@ type appContext struct {
 	addedLowFeeTicketsMSA   map[chainhash.Hash]string            // [ticket]multisigaddr
 	ignoredLowFeeTicketsMSA map[chainhash.Hash]string            // [ticket]multisigaddr
 	liveTicketsMSA          map[chainhash.Hash]string            // [ticket]multisigaddr
+	possiblyMissingTickets  map[chainhash.Hash]int64             // ticket]blockheight
```
infertux: Nitpick: missing opening bracket in comment.

JoeGruffins:
@jolan hello. Are you still working on this?

chappjc:
@JoeGruffins No, he's not on it anymore.  I suggest restarting the work with the latest stakepool architecture in mind.

PR #258, opened 2018-05-16
Correct licensing

jrick:
This project contains code governed by the MIT/X11 and ISC licenses.
The ISC license and copyrights have been added to the LICENSE file in
the root of the project, and the README has been updated to reflect
the current licensing.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015-2017 The Decred developers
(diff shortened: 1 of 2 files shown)


chappjc:
How about we relicense it under ISC instead?

jrick:
If every copyright owner is contacted and agree with the change, that is possible.

The project also uses plenty of actual ISC-licensed code copied from other projects, so in reality, it's a mixed license.

PR #260, opened 2018-05-28
Replace "deCRED" occurences with "Decred".

jrick:
The name of the currency is Decred and the logo stylization does not
change that.


chappjc:
Wanna do this for dcrdata too?  :)

PR #266, opened 2018-08-13
Repress logs for unfound ticket lookups.

jrick:
The error string created by the dcrwallet JSON-RPC gettransaction when a
transaction is not found changed in the 1.3.0 release.  This change modifies
stakepoold to repress the newly-formatted errors from being logged.

While here, include a gofmt fix from Go 1.11.


chappjc:
I suggest merging this.  It is correct.

PR #270, opened 2018-10-29
Update config.go

guotie:
spilit AdminUserIDs by ,


dajohi:
I think we should drop the comma's support.  you can specify the option multiple times.

adminuserids=user@one.com
adminuserids=other@two.com

PR #287, opened 2018-12-20
linting and cleanup

chappjc:
Remove many unused function parameters and struct fields.
Move tests for +build go1.8 into general test file.


chappjc:
Rebased.

PR #289, opened 2019-01-21
 New dbtls parameter to configure encryption for the database connection

isuldor:
dbtls is a configurable parameter to set the go-sql-driver/mysql option for tls. [1]  Feedback would be very welcome as I have little experience contributing code. I was also intending to create a guide doc for how to setup mysqld with SSL on Debian. Unfortunately, while I was able to require an ssl connection and successfully connect from a remote host using mysql-client's `--ssl-verify-server-cert=on` feature, I wasn't actually able to get stakepoold to connect using tls, even with the insecure `skip-verify` option.  So I haven't really confirmed that this works, and I'm disappointed since this would have been an alternative to setting up IPSEC to secure the database connection. But I think I've read the golang documentation right, so hopefully someone will find this useful (or maybe I'll figure out what I missed someday!).

1: https://github.com/go-sql-driver/mysql#tls


dajohi:
@brianbaligad please rebase on top of current master.

isuldor:
I'm working on this, I just need to spend some time learning to work through git rebase errors.  Thanks for the patience!

PR #295, opened 2019-02-06
Update decred modules to latest stable versions

### LINKED ISSUE #293: Needs update for latest agendas
The newest version of chainparams must be used as a build dependency so that the newest vote version is used (thereby creating votes on the current agenda).

jrick:
This update includes a new set of vote agendas and will cause votes to be
created with the newest vote versions (for mainnet, this is version 6).

**This is untested, and requires testing before merging**

Fixes #293


jrick:
@dajohi please test this out on teststakepool

Issue #296, opened 2019-02-06
Sign up page has a link to the sign up page


jholdstock:
On the sign up page, under the boxes for username and password is a link "New user?" which simply reloads the current page

PR #297, opened 2019-02-06
The Sign Up page should not link back to itself.

chappjc:
Resolving https://github.com/decred/dcrstakepool/issues/296

Also bump dates in footer and license to 2019.

### DIFF
--- a/views/auth/signup.html
+++ b/views/auth/signup.html
@@ -28,15 +28,6 @@
       <input name="passwordrepeat" type="password" class="form-control" placeholder="Password" required>
 	</div>
   </div>
-<!-- <div class="form-group" style="{{if .FlashSuccess}}display: none;{{end}}">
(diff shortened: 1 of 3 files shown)


views/auth/signup.html
```diff
@@ -50,9 +41,7 @@ <h1 style="{{if .FlashSuccess}}display: none;{{end}}">Sign Up</h1>
   </div>
 
   <div class="form-group col-sm-7 pull-right" style="{{if .FlashSuccess}}display: none;{{end}}">
-<!--    <label class="col-md-4 control-label" for="reset"></label> -->
         <p><a href="/passwordreset">Forgot your password?</a></p>
-        <p><a href="/signup">New user?</a></p>
```
chappjc: Perhaps this should be `<p><a href="/signin">Already registered?</a></p>`, but I don't think it's really necessary.

PR #298, opened 2019-02-06
Use goemail

dajohi:


alexlyp:
Looks OK overall, just wondering why you're using your fork of goemail instead of a decred fork or the normal one.

dajohi:
The go net/smtp package forces a certificate to be signed by a CA.
Some email admins use self-signed certs.  This paves the way for that,
which will be a separate commit.

PR #301, opened 2019-02-08
UserToken type and reworked token-enabled handlers

chappjc:


system/controller.go
```diff
@@ -41,3 +44,46 @@ func (controller *Controller) Parse(t *template.Template, name string, data inte
...
+	if err != nil {
+		session.AddFlash("Email verification token not valid.",
+			"passwordupdateError")
+		return token, nil, false
+	}
+
+	// Check that the token is recognized.
+	passwordReset, err := helpers.PasswordResetTokenExists(dbMap, token)
+	if err != nil {
+		session.AddFlash("Password update token not valid", "passwordupdateError")
```
chappjc: Should log the error.

system/controller.go
```diff
@@ -41,3 +44,46 @@ func (controller *Controller) Parse(t *template.Template, name string, data inte
...
+		return token, passwordReset, false
```
chappjc: Should return nil instead of passwordReset, although for non-nil errors it should be nil anyway

chappjc:
This one needs testing to be 100% sure Decrediton interaction isn't broken.

PR #302, opened 2019-02-09
 admin page: validate ticket list

chappjc:
admin page: validate ticket list

models: add `HashList`
Add `DecodeHashList` and `ValidateHashList` to decode and validate a slice
of strings meant to contain Decred hashes.

gorp: Update gorp dependency to get slice support.
Create the gorm `DbMap` with `ExpandSliceArgs=true` to use the slice expansion capability.
Implement `gorp.stringer` by `HashList` for safety and future use.
`(HashList).ToStringSlice` (`gorp.stringer`) is not used since `ticketListMapper` uses a
`[]string`, but this would break silently if it used an actual `HashList`.
Adding `ToStringSlice` prevents this possible silent breakage.


controllers/main.go
```diff
@@ -1119,73 +1130,88 @@ func (controller *MainController) AdminTicketsPost(c web.C, r *http.Request) (st
...
+
+	case "remove":
 		actionVerb = "removed"
-		query := "DELETE FROM LowFeeTicket WHERE TicketHash IN (\"" +
-			strings.Join(r.PostForm["tickets[]"], ",") + "\")"
-		_, err := dbMap.Exec(query)
+		// To use gorm's slice expansion, use a mapper with a ticketList. For
+		// three tickets in the list, gorm will expand this to:
+		//     "... IN (:Tickets0,:Tickets1,:Tickets2)".
+		// This allows each string in the list two be it's own argument.
```
degeri:
```suggestion
		// This allows each string in the list to be it's own argument.
```

chappjc: Good catch.  Also "it's" is not right here.

PR #307, opened 2019-02-13
controller: Never rescan during script imports

jrick:
The only scripts that are imported by the controller are ones that were just
created from a createmultisig RPC, and should not have any history that requires
a rescan to find previous outputs.  Remove the rescan request from the
importscript parameters to prevent the wallet (which no longer serializes and
batches rescan jobs) to prevent extraneous resource usage for unnecessary
rescans.


chappjc:
Looks like the previously merged PR didn't gofmt:

```
controllers/main.go:148: File is not `gofmt`-ed with `-s` (gofmt)
        smtpUrl := "smtp://"
```

Could you pass it through gofmt, @jrick ?

PR #308, opened 2019-02-13
multi: more error checking

dajohi:
cleanup some code while here.

### DIFF
--- a/backend/stakepoold/log.go
+++ b/backend/stakepoold/log.go
@@ -21,12 +21,7 @@ type logWriter struct{}
 
 func (logWriter) Write(p []byte) (n int, err error) {
 	os.Stdout.Write(p)
-	// Normally this if isn't required, but the tests in this main package need
-	// to work with uninitialized log rotators.
-	if logRotator != nil {
-		logRotator.Write(p)
-	}
-	return len(p), nil
+	return logRotator.Write(p)
 }
 
 // Loggers per subsystem.  A single backend logger is created and all subsytem
@@ -110,13 +105,4 @@ func setLogLevels(logLevel string) {
 	for subsystemID := range subsystemLoggers {
 		setLogLevel(subsystemID, logLevel)
 	}
-}
-
-// pickNoun returns the singular or plural form of a noun depending
-// on the count n.
-func pickNoun(n int, singular, plural string) string {
-	if n == 1 {
-		return singular
-	}
-	return plural
 }
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -87,6 +87,11 @@ type WinningTicketsForBlock struct {
 	winningTickets []*chainhash.Hash
 }
 
+const (
+	// TODO: sync with controllers/main.go MaxUsers
+	numServicePaymentFeeAddresses uint32 = 10000
+)
+
 var (
 	cfg              *config
 	errDuplicateVote = "-32603: already have transaction "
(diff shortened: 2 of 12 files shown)


backend/stakepoold/server.go
```diff
@@ -1112,23 +1116,19 @@ func (ctx *appContext) processWinningTickets(wt WinningTicketsForBlock) {
...
-				voteCfg.VoteBits = ctx.votingConfig.VoteBits
-				log.Infof("userid %v multisigaddress %v vote "+
-					"version mismatch user %v stakepoold "+
-					"%v using votebits %d",
-					voteCfg.Userid, voteCfg.MultiSigAddress,
-					voteCfg.VoteBitsVersion,
-					ctx.votingConfig.VoteVersion,
-					voteCfg.VoteBits)
-			}
+			voteCfg.VoteBits = ctx.votingConfig.VoteBits
```
chappjc: I'm wondering why we shouldn't also set `voteCfg.VoteVersion = ctx.votingConfig.VoteVersion`.  The wallets are going to vote with their versions, not the incorrectly set version.  However, it doesn't seem to make any difference since `(*appContext)vote` does even use the `VoteVersion` field for the this reason.  Really if the code gets here, dcrstakepool should schedule a `(*MainController).CheckAndResetUserVoteBits` or something like except for just this user and not all users.

controllers/dcrclient.go
```diff
@@ -622,7 +622,12 @@ func (w *walletSvrManager) executeInSequence(fn functionName, msg interface{}) i
 
 		if !w.checkForSyncness(spuirs) {
 			log.Infof("StakePoolUserInfo across wallets are not synced.  Attempting to sync now")
-			w.syncTickets(spuirs)
+			err := w.syncTickets(spuirs)
+			if err != nil {
+				log.Errorf("failed to syncTickets: %v", err)
+				resp.err = err
+				return resp
```
chappjc: I do not think we want to quit the `executeInSequence` loop (effectively kill dcrstakepool) if `syncTickets` has an error.  It will have a non-nil error if `rpcclient.(*Client).AddTicket` has an error.  It is not clear what would cause this error, but it could easily be transient connectivity issues, or an already added ticket, etc.

See the docs  for `(*walletSvrManager).StakePoolUserInfo`, the function that gets to this point in `executeInSequence`:

```go
// StakePoolUserInfo gets the voting service user information for a given user.
//
// This can race depending on what wallet is currently processing, so failures
// from this function should NOT cause fatal errors on the web server like the
// other RPC client calls.
func (w *walletSvrManager) StakePoolUserInfo(...
```

chappjc: I suggest we put this as `_ = w.syncTickets(spuirs)` with a comment, or simply log the error (from `AddTicket`).

backend/stakepoold/log.go
```diff
@@ -21,12 +21,7 @@ type logWriter struct{}
 
 func (logWriter) Write(p []byte) (n int, err error) {
 	os.Stdout.Write(p)
-	// Normally this if isn't required, but the tests in this main package need
-	// to work with uninitialized log rotators.
```
chappjc: Is this comment not accurate in this project?

backend/stakepoold/server.go
chappjc: But that can wait.  Just restarting dcrstakepool triggers this.

backend/stakepoold/server.go
```diff
@@ -253,18 +254,16 @@ func evaluateStakePoolTicket(ctx *appContext, tx *wire.MsgTx, blockHeight int32)
...
+		return nil, err
 	}
 	msgTx := wire.NewMsgTx()
 	if err = msgTx.Deserialize(bytes.NewReader(txBytes)); err != nil {
-		log.Warnf("Deserialize failed for %v: %v", txBytes, err)
-		return nil
+		return nil, err
 	}
-	return msgTx
+	return msgTx, nil
```
chappjc: I meant to do the following in `dcrdata/txhelpers.MsgTxFromHex`, but it's not important:

```go
msgTx := wire.NewMsgTx()
err := msgTx.Deserialize(hex.NewDecoder(strings.NewReader(txhex)))
return msgTx, err
```

controllers/dcrclient.go
dajohi: for now, I will just log all the missed errors

backend/stakepoold/log.go
dajohi: not anymore.  i took the logging out of the API that the _tests's check.  Without it, the tests would panic.

PR #313, opened 2019-02-21
[WIP] use captcha fork with options and faster ServeHTTP

chappjc:
It would be nice if all VSPs did not have identical captcha parameters (distortion and other image characteristics).

TODO: set from config and use the updated ServeHTTP


chappjc:
This work is abandoned.

PR #314, opened 2019-02-22
dep: update goemail

dajohi:


chappjc:
An issue with goemail that might need to be addressed first is no-tls smtp.  I'm not sure if it's handled right in goemail since you have to set InsecureSkipVerify.  You need to do this if you have a mail server or bridge on localhost:25

chappjc:
Oh, nevermind.  I'm thinking of a self-signed cert situation.  All good.

PR #319, opened 2019-03-06
Adds a Cache-Control header to stop sensitive data from being cached

### LINKED ISSUE #318: Cached user data needs to be cleared on logout
Reproduction steps:
1. Login to stakepool
2. Visit user pages like site.com/tickets or site.com/settings 
3. Click on the "Logout button" and get redirected to site.com
4. Hit the back button in the browser and observe that the user pages are still viewable. 

This is only an issue if an attacker has physical access to your system. 

degeri:
Adds ` "Cache-Control", "private,no-store,no-cache"` to html responses, this will fix #318

### DIFF
--- a/system/core.go
+++ b/system/core.go
@@ -139,7 +139,7 @@ func (application *Application) Route(controller interface{}, route string) web.
 
 		switch code {
 		case http.StatusOK, http.StatusProcessing, http.StatusServiceUnavailable:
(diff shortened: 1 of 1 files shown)


chappjc:
Good, although ideally the header would not be set this way in /stats, maybe the home page too.

PR #323, opened 2019-03-16
Delete password reset tokens on email change

### LINKED ISSUE #320: Invalidate password reset token on email change
When a user changes their email ID any password reset tokens that are valid for that user should be invalidated. 

Attack Scenario:
Lets say example1@example.com email is compromised and is in the hands of the attacker. The attacker requests for a password reset email, in the mean time legituser logs in to the VSP and changes the email to a secure email example2@example.com. Now the attacker can reset the password using the link that is available in example1@example.com and take over the account. 

This is a very **low risk** issue since the reset link invalidates in an **hour** and requires that the email of the user gets compromised (which is out of our control).

degeri:
Fixes #320 

This will delete all reset tokens of user when their email gets changed.

### DIFF
--- a/helpers/auth.go
+++ b/helpers/auth.go
@@ -21,6 +21,11 @@ func EmailChangeComplete(dbMap *gorp.DbMap, token models.UserToken) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


degeri:
Can we merge this ? @dajohi

PR #336, opened 2019-04-16
config: support multiple admin user IDs

### LINKED ISSUE #329: Setting multiple adminuserids doesnt work
From the sample config file:

```
; No default in case UserId 1 is a shared account of some sort.
;adminuserids=1

; Multiple values can be used and are separated by a comma.
;adminuserids=1,2,3
```

Setting a single ID works as expected - the user is granted admin permissions.
Setting multiple IDs does not work. None of the specified users get admin permissions.

jholdstock:
Closes #329


dajohi:
Closed by #270

PR #369, opened 2019-05-02
Show a message when no tickets to display. 

### LINKED ISSUE #352: [Design] Tickets page: Show a message when no tickets to display.
[The new design](https://github.com/decred/dcrstakepool/issues/269) introduced collapsible accordions which display a users tickets. At the moment, clicking an empty accordion seems to do nothing which is quite strange. It would be better to display some kind of message eg "No tickets to display".

jholdstock:
Closes #352

@MariaPleshkova  could you take a look a this please?

Preview - see the "Expired" tickets 
Screenshot from 2019-05-02 17-08-54

### DIFF
--- a/public/css/styles.css
+++ b/public/css/styles.css
@@ -7953,10 +7953,12 @@ input[type="submit"]:disabled {
 
 .ticket_accordion input[type='checkbox']:checked + label + .accordion__contents {
   max-height: 300px;
-  overflow-y: scroll; }
+  overflow-y: auto; }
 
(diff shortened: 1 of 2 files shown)


MariaPleshkova:
I think we can leave the same concept as when there are tickets to show, but remove an icon and change the color of the text to **#c4cbd2**

That's what I'm talking about:
image

jholdstock:
Thanks Maria, I will make that change.

jholdstock:
Done.

PR #375, opened 2019-05-08
Reset captcha value

### LINKED ISSUE #343: Reset captcha value in session after signup/passwordreset action
The current captcha implementation has some issues.

It is doing the validation only once and then it is basically whitelisting the current session to do a many signup or reset actions as they want. 

While this does stop fully non interactive bots an attacker only needs to solve the catpcha once , copy the sessions to an automated script and will be able to make unlimited requests on the reset and signup forms. 

> session.Values["CaptchaDone"] = false 

needs to be done after signup or password reset action.

degeri:
This will reset the the captcha to "false" after each sensitive request.
fixes #343

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -1390,6 +1390,8 @@ func (controller *MainController) PasswordResetPost(c web.C, r *http.Request) (s
 	if !controller.IsCaptchaDone(c) {
 		session.AddFlash("You must complete the captcha.", "passwordresetError")
 		return controller.PasswordReset(c, r)
+	} else {
+		session.Values["CaptchaDone"] = false
 	}
 
 	remoteIP := getClientIP(r, controller.realIPHeader)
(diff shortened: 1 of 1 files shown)


degeri:
Looked at it again.

We cant remove `session.Values["CaptchaDone"]` right now cause there are are actually two requests 1. Captcha solve -> 2. Sensitive action, this requires that we "track" that a captcha is solved and then invalidate it such that it can only be used once.  

For a "proper" fix we need to completely remove `c.Env["CaptchaDone"]` and `session.Values["CaptchaDone"]` and load up the captcha and the sensitive form in a single page / request (Would require some UX changes too). That needs be a whole another PR imo.

jholdstock:
Discussed further on matrix. Need to set both env and session vars. Also, inside function SettingsPost() we need to set the vars in an `else` statement, to be consistent with the other two instances. 

Should be good to merge after this.

PR #377, opened 2019-05-10
Referrer fix

### LINKED ISSUE #376: Password reset token leaking to github
While on the reset password page if the user clicks on the "The source code is available at GitHub" button the request to github.com leaks the full URL in the referrer. This is benign in most cases. But in this page it exposes the secret password reset link. 

This has a very low impact: 
1. It is leaking to github.com, the chances of an attack from github.com is small. 

2. Requires the user to do an unlikely action. 

3. The reset link will expire in an hour or after the user resets the password.

4. With the reset link an attacker will not be able to determine the user's email ID.

image

degeri:
Fixes #376  . Also fixes a minor typo "sode" => "code".

### DIFF
--- a/views/footer.html
+++ b/views/footer.html
@@ -5,7 +5,7 @@
           <div class="col-md-4 col-12 footer__credit-column">
             <div class="d-flex justify-content-center h-100">
               <div class="align-self-center">
-                <p class="mb-0">Decred developers | 2016-2019<br>The source sode is available at <a href="https://github.com/decred">GitHub</a></p>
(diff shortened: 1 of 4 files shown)


views/footer.html
```diff
@@ -5,7 +5,7 @@
           <div class="col-md-4 col-12 footer__credit-column">
             <div class="d-flex justify-content-center h-100">
               <div class="align-self-center">
-                <p class="mb-0">Decred developers | 2016-2019<br>The source sode is available at <a href="https://github.com/decred">GitHub</a></p>
+                <p class="mb-0">Decred developers | 2016-2019<br>The source code is available at <a href="https://github.com/decred" rel="noopener noreferrer">GitHub</a></p>
```
dajohi: can you append /dcrstakepool onto the github link?

Issue #227, opened 2017-10-30
remove dcrstakepool <-> dcrwallet link


jolan:
Two parts of this are tracked elsewhere:

- stakepooluserinfo on the user tickets page can be replaced with pulling stats gRPC once #219 goes in and #201 is completed. (#180)

- Stats page and API stats can be changed over from getstakeinfo to use the data from stakepoold gRPC/MySQL.   API will need to be bumped to v3 and dcrwebapi changed accordingly.  (#180)

The other parts then would be:

- port the dcrwallet script/ticket consistency checks that dcrstakepool runs on startup to stakepoold
- move importscript functionality into stakepoold

jolan:
As a continuation of this, it would probably make sense to change wallethosts and stakepooldhosts to votinghosts and webdatahosts or something.

That way a stakepoold admin could have a single 'webdatahost' and we wouldn't have to do the executeInSequence stuff and have hard requirements on anything other than 'votinghosts'.

jolan:
This would probably close out #70 and #166 too since gRPC+context is way less flimsy than rpcclient in my experience.

jolan:
I think the script/ticket consistency checks should be more "warn and continue". They could probably just run on a ticker in dcrstakepool itself and log.Warn the issues that need to be addressed.  I believe @jrick has fixed most if not all of the issues that caused the voting wallets to de-sync ever so slightly over time so the self-repair stuff can probably be dropped.  A 'rescanwallet' button on the admin interface would probably be the best replacement.

JoeGruffins:
@jholdstock 
I'll do the GetStakeInfo() rpc which is used in the stats pages and home page unless someone else is already on it.
Should this also verify that information is the same between stakepoold instances, or just return the first found?
What is say, blockheight is different? Perhaps a blockheight test should be done on start-up anyways? To make sure all wallets are synced.

jholdstock:
For now I would say to mirror whatever behaviour currently exists, and then we can add enhancements later as a separate piece of work.

PR #381, opened 2019-05-18
introduce stakepoold connection manager

jholdstock:
`dcrstakepool` connects to multiple instances of `stakepoold` and `dcrwallet` over gRPC. The code managing the `dcrwallet` gRPC connections is significantly more robust than the `stakepoold` counterpart.
`dcrwallet` connections are managed in a single place - namely [dcrclient.go](https://github.com/decred/dcrstakepool/blob/master/controllers/dcrclient.go). This file is responsible for comms to the multiple `dcrwallet` connections, error handling, retrying etc. Management of `stakepoold` connections is scattered throughout the code.

This PR introduces a connection manager for the stakepoold connections, similar to what exists for dcrwallet. Making stakepoold connection management more robust will become important in the future as we take further steps towards #227 and the `dcrwallet` connections are removed. The PR only refactors existing code and does not add any new functionality.

**_Marking the PR as a draft because I would like to properly validate the work on testnet before it is merged. Not possible right now because testnet blocks are not being mined_**

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -34,9 +34,7 @@ import (
 	"github.com/gorilla/csrf"
 	"github.com/zenazn/goji/web"
 
-	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/connectivity"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


jholdstock:
Validated on testnet, works as expected

Issue #226, opened 2017-10-30
items for clean up in dcrstakepool initialization/config


jolan:
- main loop should look more like stakepoold where we don't bother with numeric error codes
- the validation of the 2 pubkeys should probably be done in config rather than in NewMainController so we don't give confusing error messages like in #216
- enablestakepoold should be removed as an option

kiahmed:
well, I got the same error as i put xpub key of two dcrwallets one after another separated by comma . how can I put multiple voting nodes key then?

JoeGruffins:
@dajohi I believe this can be closed.

jholdstock:
Agreed. This is done

PR #391, opened 2019-05-24
Clean Up: runMain() in server.go to return error

JoeGruffins:
if at first you don't succeed...

change runMain() in server.go to run errors

this may cause some errors to be printed twice


JoeGruffins:
related to issue #226

PR #396, opened 2019-05-25
Clean Up: move pubkey parsing to config

JoeGruffins:
part of issue #226 

Move validation of pubkeys, length and network, to config.

I made the parsed keys into package level variables because it felt weird to include them in the config struct since they are not configured by the user.

### DIFF
--- a/config.go
+++ b/config.go
@@ -285,14 +285,6 @@ func (c *config) parsePubKeys(params *chaincfg.Params) error {
 	return nil
 }
 
-func (c *config) GetFeeKey() *hdkeychain.ExtendedKey {
-	return coldWalletFeeKey
-}
-
-func (c *config) GetVoteKey() *hdkeychain.ExtendedKey {
-	return votingWalletVoteKey
-}
-
 // newConfigParser returns a new command line flags parser.
 func newConfigParser(cfg *config, so *serviceOptions, options flags.Options) *flags.Parser {
 	parser := flags.NewParser(cfg, options)
--- a/config.go
+++ b/config.go
@@ -263,24 +263,24 @@ func fileExists(name string) bool {
 	return true
 }
 
-// validate pub vote and key keys as belonging to the network
+// validate pub vote and fee keys as belonging to the network
 func (c *config) parsePubKeys(params *chaincfg.Params) error {
 	// Parse the extended public key and the pool fees.
 	var err error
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -461,6 +495,13 @@ func loadConfig() (*config, []string, error) {
 		}
 	}
 
+	if err := cfg.parsePubKeys(activeNetParams.Params); err != nil {
+		str := "%s: Failed to parse Pubkeys"
+		err := fmt.Errorf(str+": %v", funcName, err)
```
jholdstock: Please avoid string concatenation here. You can instead use

```
err := fmt.Errorf("%s: Failed to parse Pubkeys: %v", funcName, err)
```

JoeGruffins: Yes sir!

I also realized the check should be made after checking whether keys are set at all, so moved it down a bit.

config.go
```diff
@@ -259,6 +263,36 @@ func fileExists(name string) bool {
...
+	if err != nil {
+		return fmt.Errorf("voting wallet extended public key: %v", err)
+	}
+	if !votingWalletVoteKey.IsForNet(params) {
+		return fmt.Errorf("voting wallet extended public key is for wrong network")
+	}
+	return nil
+}
+
+func (c *config) GetFeeKey() *hdkeychain.ExtendedKey {
```
chappjc: Style issue: https://golang.org/doc/effective_go.html#Getters
We try to follow this in general

chappjc: But these getters seem unneeded.  `coldWalletFeeKey` is already accessible by main (same package).  Also, the receiver of this getter is pointless.

JoeGruffins: Thank you Chapp! I will start reading through effective go.

I didn't even realise they were the same package.  Still learning, thank you.

Is this better?

chappjc:
OK by me.  The new `parsePubKeys` function is a candidate for a test case, but I won't make a fuss about it as long as testing on a testnet vsp/stakepool setup has been done.  @JoeGruffins have you put it through the paces on a full setup?

JoeGruffins:
@chappjc 
I checked several times to make sure the keys were being sent to controllers/main.go properly.  I also ran my testnet node for a day with this code and ticket purchases and fees were unaffected.  

But I want to write a test! You mean the testing package right? I want to do that. Give me a day.

chappjc:
Great!  Yup, some test in a new config_test.go file.

JoeGruffins:
test added

PR #398, opened 2019-05-26
remove enable stake pool as an option, always enabled

JoeGruffins:
part of issue #226 

remove ability to disable stakepoold communications.

After making these changes, the user will need to remove the "enablestakepoold=" option from their config or see an error during parsing

This will cause merge conflicts with another of my pull requests...

### DIFF
--- a/config.go
+++ b/config.go
@@ -624,6 +624,10 @@ func loadConfig() (*config, []string, error) {
 		}
 	}
 
+	if cfg.EnableStakepoold {
+		fmt.Fprintln(os.Stderr, "Option --enablestakepoold is deprecated.  Please remove from your config file")
+	}
+
 	if len(cfg.StakepooldHosts) == 0 {
 		str := "%s: stakepooldhosts is not set in config"
 		err := fmt.Errorf(str, funcName)
--- a/config.go
+++ b/config.go
@@ -624,10 +624,6 @@ func loadConfig() (*config, []string, error) {
 		}
 	}
 
-	if cfg.EnableStakepoold {
-		fmt.Fprintln(os.Stderr, "Option --enablestakepoold is deprecated.  Please remove from your config file")
-	}
-
 	if len(cfg.StakepooldHosts) == 0 {
 		str := "%s: stakepooldhosts is not set in config"
 		err := fmt.Errorf(str, funcName)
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -105,7 +105,6 @@ type config struct {
 	AdminIPs           []string `long:"adminips" description:"Expected admin host"`
 	AdminUserIDs       []string `long:"adminuserids" description:"User IDs of users who are allowed to access administrative functions."`
 	MinServers         int      `long:"minservers" description:"Minimum number of wallets connected needed to avoid errors"`
-	EnableStakepoold   bool     `long:"enablestakepoold" description:"Enable communication with stakepoold"`
```
chappjc: Since unrecognized parameters throw errors, I'm generally in favor of deprecating before removing.  That is make all the changes in this PR, except leave this line and change the description to say that something like "(deprecated. stakepoold is required)".  Then config.go can just print a warning if `enablestakepoold` is set that the user should remove the option since it's deprecated.

JoeGruffins: OK! I'll do that. Thank you.

chappjc: Please also rebase and resolve the server.go conflict.  Lemme know if you need git workflow tips.

JoeGruffins:
rebased and will not err if option is set, option is just ignored, description tells of deprecation

dajohi:
@JoeGruffins please rebase

JoeGruffins:
rebased

config.go
```diff
@@ -109,10 +109,11 @@ type config struct {
...
 	MaxVotedAge        int64    `long:"maxvotedage" description:"Maximum vote age (blocks since vote) to include in voted tickets table"`
 	Description        string   `long:"description" description:"Operators own description of their VSP"`
 	Designation        string   `long:"designation" description:"VSP designation (eg. Alpha, Bravo, etc)"`
+	//deprecated
+	EnableStakepoold bool `long:"enablestakepoold" description:"This option has been deprecated and is not settable"`
```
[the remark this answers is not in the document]
chappjc: To clarify, in config.go it needs to log a message for the user about the deprecation.  Whenever something gets deprecated, the user needs a strong warning because by definition, the option will be removed in the future and thus break the config.

config.go
```diff
@@ -624,59 +624,57 @@ func loadConfig() (*config, []string, error) {
 		}
 	}
 
-	if cfg.EnableStakepoold {
```
dajohi: May want to add something like:

```go
if cfg.EnableStakepoold {
  fmt.Fprintfln(os.Stderr, "Option --enablestakepoold is deprecated.  Please remove from your config file")
}
```

JoeGruffins: Thank you. Done.

config.go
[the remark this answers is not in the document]
JoeGruffins: srry I overlooked this comment.  A message is now logged if set true.

Issue #353, opened 2019-05-02
[Design] Tickets page: Not all voted tickets are shown


jholdstock:
In the "Voted Tickets" section of the tickets page, only tickets which have voted within the last 8640 blocks are shown. We should either show all of the users voted tickets, or explain in the UI why they are not all displayed.

jholdstock:
As of #402, the number of tickets displayed is now configurable. The default is to display the most recent 1000 tickets.

PR #402, opened 2019-06-07
Show last N voted tickets

jholdstock:
Rendering the `/tickets` page with a huge number of tickets results in the page loading very slowly.

|No. Tickets|Page Load (ms)|
|---|---|
|10|180|
|100|250|
|500|420|
|1000|600|
|2500|1300|
|5000|2200|
|10000|5000|
|17500|9600|
|25000|14900|
|37500|25200|
|50000|39000|

g

We currently mitigate this issue by only rendering the page with tickets which have voted in the last N blocks. This is not ideal for various reasons:

- The default for N is 8064, which means that in the worst possible case the page is rendered with 40,320 tickets. This means a page load of around 30 seconds.
- We need to call dcrwallet RPC to find the current best block
- A user may only have a single voted ticket, but if that ticket voted more than N blocks ago, it wont be displayed on the page.

I propose that we switch from "tickets which voted in the last N blocks" to "last N voted tickets", and I propose that a sensible default value is 1000. This provides the user with a decent amount of historic data whilst ensuring that page load time remains below well below a second.

Relevant to #353
Removes a dcrwallet RPC call, so also progress towards #227

### DIFF
--- a/config.go
+++ b/config.go
@@ -23,29 +23,29 @@ import (
 )
 
 const (
-	defaultBaseURL        = "http://127.0.0.1:8000"
-	defaultClosePoolMsg   = "The voting service is temporarily closed to new signups."
-	defaultConfigFilename = "dcrstakepool.conf"
-	defaultDataDirname    = "data"
-	defaultLogLevel       = "info"
-	defaultLogDirname     = "logs"
-	defaultLogFilename    = "dcrstakepool.log"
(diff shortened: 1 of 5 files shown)


jholdstock:
@chappjc could you take a look at this please? It's related to a discussion we had in the past.

chappjc:
Sure.  I recall this well.  Will review today

PR #404, opened 2019-06-09
tickets page: distinguish between live and immature

### LINKED ISSUE #349: [Design] Tickets page: Distinguish between live and immature
On the tickets page in [the new design](https://github.com/decred/dcrstakepool/issues/269), there is a distinction between Live and Immature tickets. This has not been implemented yet because the information was not readily available in the code.

C-ollins:
Closes #349 
Screenshot 2019-06-09 at 6 30 55 PM

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -1848,19 +1848,20 @@ type TicketInfoInvalid struct {
 	Ticket string
 }
 
-// TicketInfoImmature represents immature (mined) tickets that have yet to
+// TicketInfo represents live immature (mined) tickets that have yet to
 // be spent by either a vote or revocation.
-type TicketInfoImmature struct {
+type TicketInfo struct {
 	TicketHeight uint32
 	Ticket       string
 }
 
+// TicketInfoImmature represents immature (mined) tickets that have yet to
+// be spent by either a vote or revocation.
+type TicketInfoImmature TicketInfo
+
 // TicketInfoLive represents live tickets that have yet to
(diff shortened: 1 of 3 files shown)


controllers/main.go
```diff
@@ -1848,7 +1848,14 @@ type TicketInfoInvalid struct {
 	Ticket string
 }
 
-// TicketInfoLive represents live or immature (mined) tickets that have yet to
+// TicketInfoImmature represents immature (mined) tickets that have yet to
+// be spent by either a vote or revocation.
+type TicketInfoImmature struct {
```
jholdstock: Im not very keen on the duplication here. We already have a struct below named `TicketInfoLive` and it has the same fields as this new struct.

Perhaps we could rename the existing one something more generic, and reuse that one.

jholdstock:
Discussed with @C-ollins that a better way to achieve this would be to have dcrwallet set the status correctly to "immature" rather than "live". 

This needs to be changed because the call to GetBestBlock is removed in #402, so this change should not depend on the current height being known.

jholdstock:
Work to identify immature/live tickets being done in decred/dcrwallet#1478

dajohi:
Will need to require RPC version check
https://github.com/decred/dcrwallet/pull/1481

jholdstock:
https://github.com/decred/dcrstakepool/blob/b25f4307a2f1f956ab04c32919f7e88f66a4938d/backend/stakepoold/rpcclient.go#L15

PR #408, opened 2019-06-13
stats page: split stats & status into network stats & pool stats

### LINKED ISSUE #355: [Design] Stats page
There are various statistics which were shown in the old site but missing in the new site:
- `Difficulty` (ie ticket price)
- `TotalSubsidy`
- `UserCount`
- `UserCountActive`
- `Network`
- `PoolFees`
- `PoolStatus`
- `PoolEmail`

Also I think this page could do a better job of distinguishing personal/pool/network stats. It is not clear at the moment which is which

C-ollins:
Closes #355
Screenshot 2019-06-12 at 6 14 14 PM

### DIFF
--- a/views/stats.html
+++ b/views/stats.html
@@ -4,7 +4,7 @@
 		<div class="row mx-3">
 			<section class="block">
 				<div class="col-12 block__title">
-					<h1><span>Stats and Status</span></h1>
+					<h1><span>Pool Stats</span></h1>
 				</div>
 				<div class="col-12 mb-4">
 					<div class="row">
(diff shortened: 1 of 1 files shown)


jholdstock:
Hey @C-ollins,

This looks like a step in the right direction but I have a couple of suggestions.

- Pool status, proportion live and Network are already listed on the homepage and I dont think we need to duplicate them here.
- Total Subsidy is actually a Pool stat and not a network stat. It represents the sum of voting rewards which have been awarded to all users of this pool.
- Let's truncate ticket price to 2 decimal points.
- For headings like "Live", "Immature" etc, can we update to say "Live Tickets", "Immature Tickets" etc

PR #421, opened 2019-06-25
Remove autorefill in the captcha form

degeri:
This fixes the annoying and useless auto suggestion that come up when you try to fill a captcha.


JoeGruffins:
That is indeed annoying. After you do a few you can't see the captcha anymore. And it obviously shouldn't be found.

PR #422, opened 2019-06-25
mulit: remove voting public key from config and check fee public key on start up

### LINKED ISSUE #407: Config: Check that cold wallet key and voting wallet key are not the same
According to @dajohi there have been cases in the past where users have attempted to use the same wallet and/or account for both voting and collecting voting fees.

We can help VSP operators to avoid these kind of mistakes by adding some simple checks to the code, either check if the 2x pubkeys are identical, or check if they belong to 2x distinct wallets. The former is trivial to implement, I am not sure if RPCs exist for the latter.

JoeGruffins:
After stakepools are confirmed running, loop through them sending a grpc request to perform rpc requests on their respective dcrd to get maps of master pubkeys to account. found pairs are printed and then the user defined key values are compared to the found keys. The voting key must be found to continue, the fee key must not be found.

Still want to write some tests for the grpc and rpc methods.

closes #407


JoeGruffins:
Hmm. rpc tests are a little more than I bargained for.  It's on my todo list... I will test a little more manually.

stakepooldclient/stakepooldclient.go
```diff
@@ -352,3 +352,29 @@ func (s *StakepooldManager) RPCStatus() []string {
...
+		resp, err := client.MyMasterPubKeys(context.Background(), &pb.MyMasterPubKeysRequest{})
+		if err != nil {
+			log.Warnf("GetMasterPubkeys gRPC failed on stakepoold instance %d: %v", i, err)
+			continue
+		}
+		for _, pubkey := range resp.MasterPubKeys {
+			pubkeymap[pubkey.PubKey] = pubkey.Account
+		}
+	}
+	if len(pubkeymap) == 0 {
```
jholdstock: I dont think we need this check because afaik this condition should be impossible. Every instance of dcrwallet will always have at least one MasterPubKey because every instance of dcrwallet has a `default` account. Can anybody else confirm?

backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -354,6 +355,38 @@ func (ctx *AppContext) UpdateUserDataFromMySQL() error {
...
+		return nil, err
+	}
+	keymap := make(map[string]string)
+	for acct := range acctmap {
+		// skip imported
+		if acct == "imported" {
+			continue
+		}
+		// There is no construct for getmasterpubkey in dcrd yet so use raw json
+		rawreq := []json.RawMessage{json.RawMessage(`"` + acct + `"`)}
```
jholdstock: I'm not totally comfortable with using raw requests here. A better solution would be to get the RPC added to dcrwallet properly.

backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -354,6 +355,38 @@ func (ctx *AppContext) UpdateUserDataFromMySQL() error {
...
+		log.Errorf("MyMasterPubKeys: ListAccounts rpc failed: %v", err)
+		return nil, err
+	}
+	keymap := make(map[string]string)
+	for acct := range acctmap {
+		// skip imported
+		if acct == "imported" {
+			continue
+		}
+		// There is no construct for getmasterpubkey in dcrd yet so use raw json
```
jholdstock: This is dcrwallet, not dcrd

JoeGruffins: the methods are in the dcrd repo, but ill change that

backend/stakepoold/rpc/rpcserver/context.go
JoeGruffins: oh man, do you think I could get a commit in dcrd? hell yeah!

stakepooldclient/stakepooldclient.go
JoeGruffins: the requests don't return error if they err on rpc calls, so this was to err out if the children didn't find anything. It would be better to have this behavior in the children to be more clear I guess.

JoeGruffins: behavior moved to children, stakepoold instances

backend/stakepoold/rpc/rpcserver/context.go
JoeGruffins: awaiting merge in dcrd

JoeGruffins:
Ok. The coldwalletextpub is checked to not belong to the voting wallets.  The votingwalletextpub is found automatically. votingwalletextpub is deprecated and there is a warning if set.  It has been removed from the readme and a warning is in the config file.

JoeGruffins:
I'm not sure that this travis build error is my fault.  Google says I can close and reopen the issue to trigger a rebuild.  Trying.

README.md
```diff
@@ -233,13 +233,6 @@ $ dcrwallet --create
 $ dcrwallet
 ```
 
-- Get the master pubkey from the default account.  This will be used for
```
jholdstock: I think we should **update** rather than **remove** this piece of documentation. We can explain how the key is now derived automatically from the `default` account

backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -408,6 +408,32 @@ func (ctx *AppContext) UpdateUserDataFromMySQL() error {
...
+	}
+	keyMap := make(map[string]string)
+	for acct := range acctMap {
+		// skip imported
+		if acct == "imported" {
+			continue
+		}
+		key, err := ctx.WalletConnection.GetMasterPubkey(acct, ctx.Params)
+		if err != nil {
+			log.Errorf("MyMasterPubKeys: GetMasterPubKey rpc failed: %v", err)
```
jholdstock: I think this should probably be a fatal error which causes the RPC to return an error response immediately. I dont see any reason we should tolerate an error here. 

If there is something wrong with the way things are configured, the VSP operator should get a hard failure so they know they need to fix it straight away. The way this is currently written could make an error easy to miss

backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -408,6 +408,32 @@ func (ctx *AppContext) UpdateUserDataFromMySQL() error {
...
+			continue
+		}
+		keyMap[acct] = key.String()
+	}
+	if len(keyMap) == 0 {
```
jholdstock: When looking at the dcrd code, did you confirm whether we need to keep this check or not?

stakepooldclient/stakepooldclient.go
```diff
@@ -498,3 +498,32 @@ func (s *StakepooldManager) RPCStatus() []string {
 
 	return stakepooldPageInfo
 }
+
+// MyMasterPubKeys performs gRPC MasterPubKeys requests against all stakepoold
+// instances and returns all results fetched without errors. Returns a map of keys
```
jholdstock: Comment suggests "returns without errors" but there are several places in the code below which return errors.

I would suggest that we return an error on any failure, or any on mis-matched results. All of the hot wallets should be set up with the same seed, and so they should have the same accounts and master pubkeys. Any mismatch indicates that the VSP operator has to take some remedial action

jholdstock:
> I'm not sure that this travis build error is my fault. Google says I can close and reopen the issue to trigger a rebuild. Trying.

I didnt know you could trigger a rebuild like that. Alternatively you can just log into travis with your github ID and it should allow you to press a button which rebuilds. Not sure if you need any repo permissions to do that though...

backend/stakepoold/rpc/rpcserver/context.go
JoeGruffins: Ok. That does make sense. It also makes me feel better about not checking the length of the map later.

stakepooldclient/stakepooldclient.go
JoeGruffins: Ok!

JoeGruffins:
> I didnt know you could trigger a rebuild like that. Alternatively you can just log into travis with your github ID and it should allow you to press a button which rebuilds. Not sure if you need any repo permissions to do that though...

I just do what google tells me to do.  I should start a new religion.  Would you to like to become a Goober?

backend/stakepoold/rpc/rpcserver/context.go
JoeGruffins: This should never happen.  Since we are going to stop on any error I think it is safe to take this out.

stakepooldclient/stakepooldclient.go
JoeGruffins: ok. This allows for different names, but not for different keys with the same name.  I think this is enough, unless really think they should be exact.  The accounts for my pools have different names from testing now... As long as default is the same I think the pool will function as intended. Even changing the default should not cause a problem, as long as both are changed.

JoeGruffins:
Should this increment the api version major, since it is a breaking change?

jholdstock:
> Should this increment the api version major, since it is a breaking change?

Yes please.

stakepooldclient/stakepooldclient.go
jholdstock: Copying my comment from Matrix, to ensure I dont forget about it:

> I think we can simplify things here, I dont think we need to compare all of the master pub keys. We can just get:
>
> - pool1keys["default"]
> - pool2keys["default"]
> - pool3keys["default"]
>
> If they are all equal, return that value. If any of them are different, return an error.
> I dont think we need to worry about people renaming the default account or otherwise customising the voting wallets, these wallets should be used for a single purpose - voting for the VSP. If operators want to mess around on their wallets and do things outside of the prescribed process, they can deal with that extra complexity themselves.

jholdstock:
The latest changes look good, just the API version to increment and [one outstanding comment](https://github.com/decred/dcrstakepool/pull/422#discussion_r307262657) from me.

Needs to be rebased also, some conflicts have popped up.

JoeGruffins:
Rebased and breaking this pr into its parts. The final check will also be broken into two methods.

JoeGruffins:
@jholdstock what do you think?

README.md
```diff
@@ -233,12 +233,9 @@ $ dcrwallet --create
...
-  votingwalletextpub in dcrstakepool.conf.
+- The master extended public key for the "default" account will be derived
```
chappjc: by "derived" I think this means "retrieved" (from the wallets).

backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -408,6 +408,27 @@ func (ctx *AppContext) UpdateUserDataFromMySQL() error {
 	return nil
 }
 
+// MyMasterPubKeys returns a map of master pubkeys to their account
```
chappjc: I'd prefer to change the possessive language.  i.e. "My" and "their" could be changed to more descriptive things.  From the following code, it seems the function retrieves the xpub keys for each account, excluding imported.

In api.proto, `MyMasterPubKeys` would also be better named just `MasterPubKeys` or `WalletMasterPubKeys`.

config.go
```diff
@@ -690,6 +675,11 @@ func loadConfig() (*config, []string, error) {
 		log.Warnf(str, funcName)
 	}
 
+	if cfg.VotingWalletExtPub != "" {
+		str := "%s: Config VotingWalletExtPub is deprecated and has no effect. Please remove from your config file."
```
chappjc: I'd also explicitly state that it will be removed in a future release, even though that's the definition of deprecated, and it should be removed to avoid startup errors in the future.

sample-dcrstakepool.conf
```diff
@@ -88,6 +88,8 @@ testnet=1
 ; Specified extended public key is used to generate ticketed addresses
 ; which are combined with a user address for 1-of-2 multisig.
 ; Must be the voting wallet's masterpubkey for the default account.
+; DEPRECATED: This option has no effect. votingwalletextpub is derived
```
chappjc: Here again the word "derived" is a bit confusing.  It is already derived in the wallet, dcrstakepool just retrieves it.

server.go
```diff
@@ -95,6 +97,22 @@ func runMain() error {
 		return fmt.Errorf("Failed to initialize the smtp server: %v", err)
 	}
 
+	// get the default master extended public key
```
chappjc: dcrstakepool repo might have different standards, but we generally want comments that are written as sentences to begin with a capital letter and end with punctuation.

chappjc: Maybe also reword this as "Retrieve the extended public key for the voting wallet's "default" account."

server.go
```diff
@@ -95,6 +97,22 @@ func runMain() error {
...
+	voteKey, err := stakepooldConnMan.DefaultAccountPubKey()
+	if err != nil {
+		return fmt.Errorf("Failed to get master extended public key for the \"defualt\" account: %v", err)
```
chappjc: typo: defualt

chappjc: In most cases, `error` strings should not begin with capital letters (starting with a type name that has a cap is one of the exceptions).

server.go
```diff
@@ -95,6 +97,22 @@ func runMain() error {
...
+	}
+	if votingWalletVoteKey, err = hdkeychain.NewKeyFromString(voteKey); err != nil {
+		return fmt.Errorf("Unable to convert the voting wallet master extended public key: %v", err)
```
chappjc: "convert" is a bit confusing.  I suggest "decode" instead as that more accurately reflects what `NewKeyFromString` does.

stakepooldclient/stakepooldclient.go
```diff
@@ -498,3 +498,65 @@ func (s *StakepooldManager) RPCStatus() []string {
 
 	return stakepooldPageInfo
 }
+
+// comaprePubkeys performs fn on all key, value pairs and returns
+// whether any are true
```
chappjc: Should also mention what's going on with the gRPC connections.

stakepooldclient/stakepooldclient.go
```diff
@@ -498,3 +498,65 @@ func (s *StakepooldManager) RPCStatus() []string {
...
+			}
+		}
+	}
+	return false, nil
+}
+
+// WalletsHavePubKey checks whether pubkey exists on any stakepoold
+// wallets
+func (s *StakepooldManager) WalletsHavePubKey(pubkey string) (bool, error) {
+	hasPubkeyFn := func(k string, v string) bool {
```
chappjc: `func(_ string, v string)`

stakepooldclient/stakepooldclient.go
```diff
@@ -498,3 +498,65 @@ func (s *StakepooldManager) RPCStatus() []string {
...
+	}
+	hasPubKey, err := s.comparePubKeys(hasPubkeyFn)
+	if err != nil {
+		return false, err
+	}
+	return hasPubKey, nil
+}
+
+// DefaultAccountPubKey gets the master extended public key for the
+// default account. Returns an error if all pubkeys are not the same.
```
chappjc: The same as what? Need to mention this is comparing these pubkeys between each wallet.

stakepooldclient/stakepooldclient.go
```diff
@@ -498,3 +498,65 @@ func (s *StakepooldManager) RPCStatus() []string {
...
+func (s *StakepooldManager) DefaultAccountPubKey() (string, error) {
+	var pubkey string
+	keysDontMatchFn := func(k string, v string) bool {
+		log.Infof("found master extended pubkey %s for account \"%s\"", v, k)
+		// ignore other accounts
+		if k != "default" {
+			return false
+		}
+		//first pass sets pubkey
```
chappjc: please check comment formatting throughout

stakepooldclient/stakepooldclient.go
```diff
@@ -498,3 +498,65 @@ func (s *StakepooldManager) RPCStatus() []string {
...
+	}
+	return hasPubKey, nil
+}
+
+// DefaultAccountPubKey gets the master extended public key for the
+// default account. Returns an error if all pubkeys are not the same.
+func (s *StakepooldManager) DefaultAccountPubKey() (string, error) {
+	var pubkey string
+	keysDontMatchFn := func(k string, v string) bool {
+		log.Infof("found master extended pubkey %s for account \"%s\"", v, k)
```
chappjc: Debugf

stakepooldclient/stakepooldclient.go
```diff
@@ -498,3 +498,65 @@ func (s *StakepooldManager) RPCStatus() []string {
...
+			pubkey = v
+			return false
+		}
+		// all other passes must equal
+		if pubkey != v {
+			return true
+		}
+		return false
+	}
+	nonMatch, err := s.comparePubKeys(keysDontMatchFn)
```
chappjc: I think I'd find `mismatch` to be clearer than `nonMatch`.

server.go
jholdstock: No reason for this repo to be any different, we should standardise across all (when it is reasonable to do so)

JoeGruffins:
@chappjc thank you for the review and sorry about the basic formatting mistakes.  They won't happen again.

chappjc:
Oh, no worries.  It's small stuff.

backend/stakepoold/rpc/rpcserver/server.go
```diff
@@ -243,3 +243,21 @@ func (s *stakepooldServer) ValidateAddress(ctx context.Context, req *pb.Validate
 		PubKeyAddr: response.PubKeyAddr,
 	}, nil
 }
+
+// WalletMasterPubKeys performs the gRPC command WalletMasterPubKeys.
```
JoeGruffins: Maybe I don't need this comment? I see dcrwallet doesn't comment these. 
https://github.com/decred/dcrwallet/blob/master/rpc/rpcserver/server.go

JoeGruffins: I took it out.

config.go
```diff
@@ -690,6 +675,11 @@ func loadConfig() (*config, []string, error) {
 		log.Warnf(str, funcName)
 	}
 
+	if cfg.VotingWalletExtPub != "" {
+		str := "%s: Config VotingWalletExtPub is deprecated and has no effect. Please remove from your config file. This will not be an option in a future release of dcrstakepool, and	setting it will cause a startup error."
```
jholdstock: There is a tab instead of a space between "and" and "setting"

config.go
```diff
@@ -105,7 +105,7 @@ type config struct {
 	WalletUsers        []string `long:"walletusers" description:"Usernames for wallet servers"`
 	WalletPasswords    []string `long:"walletpasswords" description:"Passwords for wallet servers"`
 	WalletCerts        []string `long:"walletcerts" description:"Certificate paths for wallet servers"`
-	VotingWalletExtPub string   `long:"votingwalletextpub" description:"The extended public key of the default account of the voting wallet"`
+	VotingWalletExtPub string   `long:"votingwalletextpub" description:"Deprecated: Do not use. This will not be an option in a future release of dcrstakepool, and	setting it will cause a startup error."`
```
jholdstock: There is a tab instead of a space between "and" and "setting"

JoeGruffins:
Will rebase after release.

stakepooldclient/stakepooldclient.go
```diff
@@ -664,3 +663,68 @@ func (s *StakepooldManager) GetStakeInfo() (*pb.GetStakeInfoResponse, error) {
...
+	return hasPubKey, nil
+}
+
+// DefaultAccountPubKey sends gRPC commands to all stakepoold and returns the
+// master extended public key for the wallet's "default" account. Returns an
+// error if pubkeys for the "default" account are not the same accross all
+// stakepoold wallets or any gRPC commands fail.
+func (s *StakepooldManager) DefaultAccountPubKey() (string, error) {
+	var pubkey string
+	keysDontMatchFn := func(k string, v string) bool {
```
itswisdomagain: Could rename parameters to `acctName, pubKey string`

stakepooldclient/stakepooldclient.go
```diff
@@ -664,3 +663,68 @@ func (s *StakepooldManager) GetStakeInfo() (*pb.GetStakeInfoResponse, error) {
...
+	}
+	return hasPubKey, nil
+}
+
+// DefaultAccountPubKey sends gRPC commands to all stakepoold and returns the
+// master extended public key for the wallet's "default" account. Returns an
+// error if pubkeys for the "default" account are not the same accross all
+// stakepoold wallets or any gRPC commands fail.
+func (s *StakepooldManager) DefaultAccountPubKey() (string, error) {
+	var pubkey string
```
itswisdomagain: Could rename to `validDefaultAccountPubKey` to signify that
- the pubkey of interest is for the default account (should be obvious but could become confusing if my previous suggestion is applied)
- the value stored here at the end of the day ought to be the same for all stakepoold instances (hence the `valid` prefix)

Could probably just use named return variables as well.

stakepooldclient/stakepooldclient.go
```diff
@@ -664,3 +663,68 @@ func (s *StakepooldManager) GetStakeInfo() (*pb.GetStakeInfoResponse, error) {
...
+func (s *StakepooldManager) comparePubKeys(fn func(string, string) bool) (bool, error) {
+	for i, conn := range s.grpcConnections {
+		client := pb.NewStakepooldServiceClient(conn)
+		resp, err := client.WalletMasterPubKeys(context.Background(), &pb.WalletMasterPubKeysRequest{})
+		if err != nil {
+			return false, fmt.Errorf("GetMasterPubkeys gRPC failed on stakepoold instance %d: %v", i, err)
+		}
+		for _, v := range resp.MasterPubKeys {
+			if fn(v.Account, v.PubKey) {
+				return true, nil
```
itswisdomagain: This return statement in this loop implies that the grpc request may likely not be made to ALL stakepoold instances. Not sure if this was intended, PR description says `After stakepools are confirmed running, loop through them sending a grpc request to perform rpc requests on their respective dcrd to get maps of master pubkeys to account`; which I read to mean perform the grpc request on ALL stakepoold instances.

It's also important to check all instances to confirm/verify that the wallets on each backend server were all restored from the same seed; i.e. checking that the master pub key is the same across all the backend server wallets.

stakepooldclient/stakepooldclient.go
```diff
@@ -664,3 +663,68 @@ func (s *StakepooldManager) GetStakeInfo() (*pb.GetStakeInfoResponse, error) {
...
+			}
+		}
+	}
+	return false, nil
+}
+
+// WalletsHavePubKey sends gRPC commands to all stakepoold and returns whether
+// pubkey exists on any wallets. Returns an error if any gRPC commands fail.
+func (s *StakepooldManager) WalletsHavePubKey(pubkey string) (bool, error) {
```
itswisdomagain: I see that this is only used to confirm that the `coldwalletextpub` does not belong to the voting wallets. I think this check can be done at the same time as the master pub key for the default account is read to prevent performing the stakepoold `WalletMasterPubKeys` grpc request more than once.

1 idea is to invoke that grpc method once for ALL instances, ensuring that all the stakepoold instances return the same result (equal number of account-key pairs, with the same key value for each account on each backend server). Then the vetted account-key map is parsed to retrieve the default account pub key and also checked to ensure it does not contain the `coldwalletextpub`.

JoeGruffins: I would rather keep the functions separate, for easier testing and such, but you are right that we only need to call the grpc once.  Because this operation only ever happens on start-up, I didn't think it was necessary, but we could cache the results, or use a clever closure scheme.

JoeGruffins: If it's all the same, I've already had numerous reviews on this pr, and it's working ok.  Unless you think the gRPC commands will need to change, I'd rather further changes be made in another pr. That alright @itswisdomagain ?

itswisdomagain: Sure, haven't tested myself but if others have tested and it works ok, further changes can be made in a subsequent PR.

stakepooldclient/stakepooldclient.go
JoeGruffins: The first true is sufficient for what I use it for. Return true if a default xpub is different and return true if xpub is found. Once a true if found there is no need to continue, and if no trues then all accounts on all wallets have been iterated.

It is checking that the master pub key for the default account is the same across all servers.

JoeGruffins: I wrote it like this to avoid iterating the map twice. Perhaps the complexity is not worth it?

stakepooldclient/stakepooldclient.go
JoeGruffins: jholdstock isn't a fan of naked returns.

stakepooldclient/stakepooldclient.go
JoeGruffins: that's true. I may make this change.

stakepooldclient/stakepooldclient.go
itswisdomagain: Could just rename the variable then

stakepooldclient/stakepooldclient.go
itswisdomagain: Don't really quite get you. What I mean to confirm is this:
Does this code retrieve the voting wallet xpub key from ALL stakepoold instances and confirm that the xpub key is the same across ALL stakepoold instances?

JoeGruffins: Yes.

It should. Does it not? 😬

stakepooldclient/stakepooldclient.go
jholdstock: Named return variables (as suggested by itswisdomagain) are fine. I actually kind of wish naming return values was enforced and not optional.

It's just naked returns which are icky.

JoeGruffins: I was thinking of it as a package deal, but you mean naming them, and then placing their names in the return I guess. Never even thought about that.

@itswisdomagain sorry, my naked comments on your pr are in the wrong place.

backend/stakepoold/rpc/rpcserver/server.go
```diff
@@ -33,8 +33,8 @@ const (
 	// collection cycle to also trigger a timeout but the current allocation
 	// pattern of stakepoold is not known to cause such conditions at this time.
 	GRPCCommandTimeout = time.Millisecond * 100
-	semverString       = "8.0.0"
-	semverMajor        = 8
+	semverString       = "9.0.0"
```
jholdstock: Needs to be updated to 9.1.0. We incremented to 9.0.0 in #526, and this PR adds a new method without breaking backwards compatability.

JoeGruffins:
I think I will close this because I think this change will ultimately become pointless.  It's not worth the time to rebase and review.

PR #427, opened 2019-06-28
Rework "Connect to Wallet" Page

### LINKED ISSUE #415: Show redeem script on connect to wallet page
Only showing the redeem script on the Tickets page is strange - the proper place for it is the Connect to Wallet page.

We should make it clear that if a user needs to re-create their wallet from seed in decrediton, they will need to manually import the redeem script after the wallet is initialised.

Edit: P2SH Address should probably move as well

### LINKED ISSUE #416: API key should be shown even after initial Decrediton connection
API key may be needed again if the user restores their Decrediton wallet from seed.

jholdstock:
- Moved "i" icon which explains the submitted public key address. It is now found next to the pubkey, rather than next to the page header.
- Always show Decrediton API key (Closes #416)
- Moved "Ticket Information" section from the Tickets page into the Connect to Wallet page (Closes #415)
- Remove duplicate code which validates addresses submitted through the Connect to Wallet page

I suggest reviewing with "Ignore whitespace changes"

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -685,21 +685,28 @@ func (controller *MainController) Address(c web.C, r *http.Request) (string, int
 
 func validateUserPubKeyAddr(pubKeyAddr string) (dcrutil.Address, error) {
 	if len(pubKeyAddr) < 40 {
-		return nil, errors.New("Address is too short")
+		str := "Address is too short"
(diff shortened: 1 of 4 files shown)


jholdstock:
Rebased.

PR #428, opened 2019-06-29
middleware: bad check

JoeGruffins:
Setting interface value to nil does not remove it as a map item.

This check is nearly always true as we often set session.Values["UserId"] = nil in the code, but this does not remove the interface from the map.  you can see this behaviour demonstrated here https://play.golang.org/p/WfodUJh2U6u

I believe the intent of this function was not to fire if there is no user id.  Right now it is searching for nothing anyway when there is no user id.


jholdstock:
PS. I discovered #430 whilst reviewing this

PR #429, opened 2019-06-29
middleware: don't use middleware when loading images

JoeGruffins:
check url for gets directed at "/assets" and "/captchas" and ignore them
in the middleware

before this change, open home, fmt.Println("ApplyDbMap") in middleware.go func ApplyDbMap
before

after this change, open home with the same print in the same place
after
I don't believe middleware should be applied to gets to pics for performance reasons, as they aren't using the information, I don't think

### DIFF
--- a/server.go
+++ b/server.go
@@ -170,6 +170,8 @@ func runMain() error {
 
 	// HTML routes
 	html := web.New()
+	// static routes
+	static := web.New()
 
 	// Execute various middleware functions.  The order is very important
 	// as each function establishes part of the application environment/context
@@ -181,10 +183,8 @@ func runMain() error {
 	html.Use(csrf.Protect([]byte(cfg.APISecret), csrf.Secure(cfg.CookieSecure)))
 
 	// Setup static files
-	html.Get("/assets/*", http.StripPrefix("/assets/",
+	static.Get("/assets/*", http.StripPrefix("/assets/",
 		http.FileServer(http.Dir(cfg.PublicPath))))
-	html.Get("/robots.txt", http.FileServer(http.Dir(cfg.PublicPath)))
-	html.Get("/favicon.ico", http.FileServer(http.Dir(cfg.PublicPath+"/images")))
 
 	// Home page
 	html.Get("/", application.Route(controller, "Index"))
@@ -229,7 +229,7 @@ func runMain() error {
 	html.Post("/register", application.Route(controller, "RegisterPost"))
 
 	// Captcha
-	html.Get("/captchas/*", controller.CaptchaServe)
+	static.Get("/captchas/*", controller.CaptchaServe)
 	html.Post("/verifyhuman", controller.CaptchaVerify)
 
 	// Stats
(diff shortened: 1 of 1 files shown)


JoeGruffins:
Actually, I think I have a better answer as soon as I figure out the correct negated regex expression.

jholdstock:
Maintaining a regex here is not appropriate because it will require changes every time the application routes change, which is going to be very error-prone and non-intuitive for new devs.

This can be achieved by simply creating a new sub-router for static files, and don't attach any middleware to that subrouter. 

I actually implemented this change some time ago, but it seems like it never made it into master. It must be lost on my local machine somewhere, I will try to find it.

JoeGruffins:
ok! That's what I'm trying to do, but then the regex wouldn't let me do not, and I'm not sure how to specify the folders easily otherwise, yet.

jholdstock:
https://github.com/decred/dcrstakepool/compare/master...jholdstock:routers?expand=1

Found it - see the changes here. I think this achieves the desired result.

jholdstock:
Note that `/robots.txt` and `/favicon.ico` are not required, they do not actually exist at these locations

JoeGruffins:
thanks, I just needed to change the order and goji does the rest. Ill remove those paths then?

server.go
```diff
@@ -246,6 +246,8 @@ func runMain() error {
 	html.Get("/logout", application.Route(controller, "Logout"))
 
 	app.Handle("/api/*", api)
+	app.Handle("/assets/*", static)
```
jholdstock: Adding the `/assets` endpoint to the `app` router will associate all of the middleware with that endpoint.

See how I used a `parent` router with no middleware to solve this problem

JoeGruffins: Did I get it this time?

isuldor:
This is a good improvement.  I just noticed dcrdata defines an nginx `location` to serve static resources directly:

```
# Serve static resources directly.
location ~ ^/(css|js|images|fonts|dist|index.js) {
        gzip_static on; # use .gz files for pre-compressed data
        root /opt/dcrdata/public;
        # Set the Cache-Control and Expires headers for the static assets.
        expires modified 2d;
}
```

chappjc:
That's a solution I would encourage, especially since nginx is so effective at handling cached static resources.

JoeGruffins:
like here ? #433

chappjc:
Oh, good stuff.  ;)

JoeGruffins:
I think #433 is missing an opening bracket.

JoeGruffins:
hey, thanks guyz, quick question. @chappjc  @isuldor 

So with this in the nginx conf:

        # Serve static resources directly.
        location /assets/*
                gzip_static on; # use .gz files for pre-compressed data
                root /opt/dcrstakepool/public;
                # Set the Cache-Control and Expires headers for the static assets.
                expires modified 2d;
        }

Should this pr we are commenting on now not be necessary? I can't tell that the above makes any difference. Even adding the bracket after location /assets/*

JoeGruffins:
Actually, I think I get it now.  Thank you!

PR #433, opened 2019-07-01
Seperate nginx config for static files

### LINKED ISSUE #397: 503 issues in new design
I noticed on https://dcrpool.dittrex.com/ there are many 503 errors returned when the application is requesting web resources from the server (eg. images, css, js).

@JoeGruffins noticed the same thing on his pool, and he tracked it down to nginx rate limiting. The provided sample nginx config has the following rate limit settings:

https://github.com/decred/dcrstakepool/blob/32d75352ba2290c33767151205a2de6e26df7b15/sample-nginx.conf#L40
https://github.com/decred/dcrstakepool/blob/32d75352ba2290c33767151205a2de6e26df7b15/sample-nginx.conf#L58

A burst limit of 5 was acceptable with the old design because it made minimal use of images etc. The homepage of the new design currently makes 25 requests on first load.

 I suggest we increase the burst limit to 50 so we can easily accommodate the new design, and have some space to grow in the future without this issue re-appearing.

jholdstock:
Closes #397

### DIFF
--- a/sample-nginx.conf
+++ b/sample-nginx.conf
@@ -53,6 +53,15 @@ http {
         ssl_prefer_server_ciphers       on;
         add_header                      Strict-Transport-Security max-age=15552001;
 
+        # Serve static resources directly.
+        location /assets/*
+                gzip_static on; # use .gz files for pre-compressed data
+                root /opt/dcrstakepool/public;
+                # Set the Cache-Control and Expires headers for the static assets.
+                expires modified 2d;
+        }
+
+        # Everything else is proxied to dcrstakepool.
         location / {
             # apply rate limiting
             limit_req zone=stakepool burst=5;


sample-nginx.conf
```diff
@@ -53,6 +53,15 @@ http {
         ssl_prefer_server_ciphers       on;
         add_header                      Strict-Transport-Security max-age=15552001;
 
+        # Serve static resources directly.
+        location /assets/*
```
JoeGruffins: Maybe there needs to be an opening bracket here.

JoeGruffins: and I think it needs a tilde for the regex to work right? But then after it works it seems I cannot use the menu on mobile.  My phone is pretty busted up thought, so I'm not sure.

`location ~ ^/assets/*`

jholdstock: Well spotted. I opened #443 to fix this.

I tested this change on my own server but I think I forgot to copy the extra bracket back into the PR - oops!

Tilde is not necessary. Tilde forces the Regex match to be case-sensitive, which we do not need here.

PR #436, opened 2019-07-14
go.mod: update to protogen version 3

JoeGruffins:
This updates the golang protobuff package to use version 3

https://github.com/golang/protobuf/releases

When I first tried to use the .regen script to make our api.pb.go I was confused by the created go file having the constant const _ = proto.ProtoPackageIsVersion3 and finally found the answer to be downgrading my golang protoc version to v1.2.0. I believe we discussed this @jholdstock ?  
This pr will allow us to use the latest protoc.  I have compiled and bought some tickets with this, without updating stakepoold, and it seems to be going fine.  I will test some more to make sure there are no problems with updating this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,13 @@ absolute path.
 If you are modifying templates, sending the USR1 signal to the dcrstakepool
 process will trigger a template reload.
 
+### Protoc
+In order to regenerate the api.pb.go file, for the gRPC connection with
+stakepoold, the following are required:
+
+- libprotoc 3.0.0 (3.4.0 is recommended)
+- protoc-gen-go 1.3.0 (1.3.2 is recommended)
+
 ## Operations
 
 - dcrstakepool will connect to the database or error out if it cannot do so.
--- a/backend/stakepoold/rpc/stakepoolrpc/api.pb.go
+++ b/backend/stakepoold/rpc/stakepoolrpc/api.pb.go
@@ -3,13 +3,14 @@
 
 package stakepoolrpc
 
-import proto "github.com/golang/protobuf/proto"
-import fmt "fmt"
-import math "math"
-
 import (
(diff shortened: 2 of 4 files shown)


jholdstock:
This is working well for me using

- go 1.12.7
- libprotoc 3.4.0
- protoc-gen-go 1.3.2

I'd be happy to merge this. My only recommendation would be to update the requirements section of README.md to indicate suggested versions of the above tools.

JoeGruffins:
Does the user need to worry about libprotoc and protoc-gen-go? I think it's only if you need to regenerate the api.pg.go file, which a user doesn't need to do.  Shall we up the recommended go version? I'm of the mind that newer is better.

jholdstock:
Hmm.. interesting point - the current README is a mix of developer information and operator information. 

I want to help others avoid the situation you encountered (incorrect tool versions) and adding the versions to the README seems like a reasonable approach. Maybe we could add this to the [stakepoold README](https://github.com/decred/dcrstakepool/tree/master/backend/stakepoold) instead?


We should definitely update the main README to go version 1.12, but lets do that as a separate task.

JoeGruffins:
I like it under Development, what do you think? I'm not entirely sure those minimal versions are correct. I will double check.
I didn't realized there was a stakepoold readme D:  I will add to it too.

jholdstock:
LGTM

JoeGruffins:
Actually, if the front page readme is fine, I don't think there's a need to put it in the stakepoold readme too.  Just more to update later.  It's also easy to miss that readme.  The tickets I bought are already voting and there doesn't seem to be any problems with communication.  I'm fairly certain the minimums are correct for version 3's of each.

dajohi:
@JoeGruffins rebase needed

JoeGruffins:
@dajohi ok!

PR #441, opened 2019-07-18
Fix issue: CNTL: error sending verification email tls: either ServerN…

dema501:
please, review my tiny patch

### DIFF
--- a/email/email.go
+++ b/email/email.go
@@ -7,6 +7,7 @@ package email
 import (
 	"crypto/tls"
 	"fmt"
+	"net"
 
 	"github.com/dajohi/goemail"
 )
@@ -33,7 +34,9 @@ func NewSender(smtpHost string, smtpUsername string, smtpPassword string,
 	}
 	smtpURL += smtpHost
 
-	tlsConfig := tls.Config{}
+	host, _, _ := net.SplitHostPort(smtpHost)
+	tlsConfig := tls.Config{ServerName: host}
+
 	smtpServer, err := goemail.NewSMTP(smtpURL, &tlsConfig)
 	if err != nil {
 		return Sender{}, err
--- a/email/email.go
+++ b/email/email.go
@@ -34,7 +34,11 @@ func NewSender(smtpHost string, smtpUsername string, smtpPassword string,
 	}
 	smtpURL += smtpHost
 
-	host, _, _ := net.SplitHostPort(smtpHost)
+	host, _, err := net.SplitHostPort(smtpHost)
+	if err != nil {
+		return Sender{}, err
+	}
+
 	tlsConfig := tls.Config{ServerName: host}
 
 	smtpServer, err := goemail.NewSMTP(smtpURL, &tlsConfig)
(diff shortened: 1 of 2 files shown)


email/email.go
```diff
@@ -33,7 +34,9 @@ func NewSender(smtpHost string, smtpUsername string, smtpPassword string,
 	}
 	smtpURL += smtpHost
 
-	tlsConfig := tls.Config{}
+	host, _, _ := net.SplitHostPort(smtpHost)
```
JoeGruffins: I think it would be better not to ignore errors. Maybe log it and if no error set the ServerName.

email/email_test.go
```diff
@@ -0,0 +1,40 @@
...
+	UseSMTPS          bool   `long:"usesmtps" description:"Connect to the SMTP server using smtps."`
+	SMTPSkipTLSVerify bool   `long:"smtpskiptlsverify" description:"Skip TLS verification when it establish connection to the SMTP server."`
+}
+
+func TestSMTPSendMail(t *testing.T) {
+	cfg := config{
+		SMTPHost:     os.Getenv("SMTPHost"),
+		SMTPUsername: os.Getenv("SMTPUsername"),
+		SMTPPassword: os.Getenv("SMTPPassword"),
+		SMTPFrom:     os.Getenv("SMTPFrom"),
```
JoeGruffins: None of these environment variables are set by default for me.  Should I be setting them before the test?

dema501: this test will be skipped, if don't to provide real credentials.
to run it 
```
SMTPHost="smtp.fastmail.com:587" SMTPUsername="**"  SMTPPassword="**" SMTPFrom="**" go test github.com/decred/dcrstakepool/email -run TestSMTPSendMail
```

JoeGruffins: I see. Maybe a comment explaining that is warranted.

email/email.go
dema501: well, in worse case `net.SplitHostPort` will return  `smtpHost` as is
https://play.golang.org/p/etsLIZv_Oz8

but you're right providing smtpHost without port it wouldn't work anyway
so, error checking will prevent it

JoeGruffins:
Thanks!

The code itself looks ok to me, but unfortunately I've yet to get smtp working with tls at all on my development environment, so I am unable to test. :(  I think we are trying to get away from users and email in the end, so I've not spent any time on trying to figure it out, but hopefully one of the real devs can give this pr the look over it deserves.

jholdstock:
I suggest @dajohi takes a look at this one, he has done some work around the email code and also maintains https://github.com/dajohi/goemail

dajohi:
@dema501 @jholdstock 

care to review?
https://github.com/dajohi/goemail/pull/7

dajohi:
@dema501 could you upgrade goemail to v1.0.1 in go.mod?

dema501:
> @dema501 @jholdstock
> 
> care to review?
> [dajohi/goemail#7](https://github.com/dajohi/goemail/pull/7)

yes, it works for me, no need to apply my patch

jholdstock:
Nice, seems like we can close this PR then. Thanks for raising it anyway dema501!

PR #443, opened 2019-07-20
Add missing bracket to nginx config

jholdstock:


JoeGruffins:
@jholdstock Have you tried using on mobile with this? I don't think you can press the menu anymore. Also, this makes the changes involving assets useless on #410 Should assets route be removed from the handler completely after this?

jholdstock:
#429 is not useless - those middleware changes are still relevant when working in dev and allowing the dcrstakepool to serve the static assets. Nginx is for production.

I have not tested this with a mobile client and I am not easily able to right now. Can you investigate the issue further? Consider that a difference between mobile and desktop experience could be caused by the difficulty of clearing the cache on a mobile browser vs desktop.

JoeGruffins:
Ok.
It's working fine now.  I think the problem was with my phone. Wouldn't be the first time. Sorry for the false alarm.

PR #445, opened 2019-07-22
main.go: add status to voting page

### LINKED ISSUE #348: [Design] Voting page: Show status of vote
[The new design](https://github.com/decred/dcrstakepool/issues/269) includes a status indicator for current voting agendas (Upcoming/In Progress/Passed/Failed). This has not been implemented because the code does not currently have access to this information. It could possibly be retrieved from voting.decred.org. If we get information from voting.decred.org, we could also retreive the long text description and display that also

JoeGruffins:
Getting status from dcrdata.org.

I'm not sure that this is the design you want.  I found it somewhere.  It always has the dots I guess, is that how it should be?

status

closes #348


jholdstock:
I have taken the following from https://voting.decred.org

Screenshot from 2019-07-24 10-15-34

However, these don't look great on a white background

Screenshot from 2019-07-24 10-20-05

I believe we want something similar for this VSP voting page. @rmykolas @MariaPleshkova do either of you have a mock-up or input for this?

MariaPleshkova:
We have almost the same ones on pi-redesign. And I think they look ok on a white bg, no?
image

JoeGruffins:
where... can I find these?

jholdstock:
I took these from https://voting.decred.org/
You can see the other styles by changing the css class `finished`. Possible values are `upcoming`, `in-progress` and `failed`.
 
Screenshot from 2019-07-29 11-13-54

jholdstock:
Alternatively you can view the politeia redesign mentioned by Maria here: https://pi-redesign.authit.co

And here is an xd for it: https://xd.adobe.com/spec/25f9f648-8d13-446f-75aa-242b21116f87-53c8/screen/1573805f-4311-4015-9d55-fba324dcbe85/Proposal-details-Desktop-1

JoeGruffins:
The statuses look like this now, copied from dcrvotingweb. What do you think?

キャプチャ

JoeGruffins:
Ok

キャプチャ

MariaPleshkova:
@JoeGruffins the statuses look ok to me.

controllers/main.go
```diff
@@ -164,6 +167,20 @@ func (controller *MainController) getNetworkName() string {
 	return controller.params.Name
 }
 
+// gets json data for current agendas from dcrdata.org
+func dcrDataAgendas() (agendasInfos []*dcrdatatypes.AgendasInfo, err error) {
+	resp, err := http.Get("https://explorer.dcrdata.org/api/agendas")
```
jholdstock: Need to switch the URL for testnet.

We can use something like:

```
"https://" + controller.getNetworkName() + ".dcrdata.org"
```

controllers/main.go
```diff
@@ -1998,14 +2015,36 @@ func (controller *MainController) Voting(c web.C, r *http.Request) (string, int)
 
 	t := controller.GetTemplate(c)
 
+	type agenda struct {
+		Agenda chaincfg.ConsensusDeployment
+		Status string
+	}
+	agendaInfos, err := dcrDataAgendas()
+	if err != nil {
+		log.Warnf("unable to retrieve data from dcrdata.org: ", err)
```
jholdstock: I would recommend constructing this error message inside dcrDataAgendas(). That way, we can easily include the full, actual URL which was requested

JoeGruffins:
I'm having a problem with rebasing. I think it will work after rpc/v4. It complains about dcrjson right now. I'll get the suggested changes in tho.

controllers/main.go
JoeGruffins: I moved the url outside, if that's also ok.

controllers/main.go
JoeGruffins: looks working!

JoeGruffins:
@jholdstock rebased!

views/voting.html
```diff
@@ -43,20 +43,40 @@ <h1><span>Voting Preferences (v{{.VoteVersion}})</span></h1>
 						<label for="accordion-control-{{$i}}">
 							<div class="row accordion__toggle">
 								<div class="col-6">
-									<h2>{{$data.Vote.Id}}</h2>
+									<h2>{{$data.Agenda.Vote.Id}}</h2>
 								</div>
+								{{if ne $data.Status ""}}
```
jholdstock: Rather than comparing to zero-value here, we can just use 
```
{{with $data.Status}}
```

views/voting.html
```diff
@@ -43,20 +43,40 @@ <h1><span>Voting Preferences (v{{.VoteVersion}})</span></h1>
...
+								<div class="col-6 text-right">
+									<!-- labels -->
+									{{if eq $data.Status "upcoming"}}
+									<div class="indicator upcoming">upcoming</div>
+									{{end}}
+									{{if eq $data.Status "in progress"}}
+									<div class="in-progress indicator">in-progress</div>
```
jholdstock: Lets be consistent with the hyphens and the caps here. E.G. we currently have "in-progress" and "Locked In".

I think css class names should be lower case and hyphenated, text displayed to the user should be title case with spaces.

JoeGruffins:
I changed the naked to a normal return to be more like other repos.

Titlecase:
titlecase

JoeGruffins:
travis saved me. I was ignoring an error after that change. oops.

PR #447, opened 2019-07-23
rpc: increment api version

jholdstock:
This is an overdue change. The API version should have been incremented when new calls were added for #227


kiahmed:
@jholdstock  the rpc changes how's that compatible with the front end now? tried a push today its breaking

jholdstock:
You need to rebuild both dcrstakepool and stakepoold

```
go install -v .
```

```
cd backend/stakepoold
go install -v .
```

kiahmed:
thanks , but already fixed it . it wasnt merged , i was pulling before it did

PR #457, opened 2019-07-31
dcrstakepool: correct dcrwallet/stakepoold count check, allow empty smtphost configuration

itswisdomagain:
- Use `config.MinServers` parameter to check the number of required dcrwallet and dcrstakepoold instances.
- Readme says `SMTPHost now defaults to an empty string so a voting service can be used for development or testing purposes without a configured mail server.`
This doesn't work currently, producing the error message instead: `Failed to initialize the smtp server: invalid smtpfrom address ""`. Support for empty `smpthost` config value is implemented in this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 config.toml
 dcrstakepool
-.vendor/
+vendor/
 .vscode/
+.idea/
 controllers/config.go*
 testing/
 *.orig
--- a/config.go
+++ b/config.go
@@ -42,7 +42,6 @@ const (
 	defaultPublicPath      = "public"
 	defaultTemplatePath    = "views"
 	defaultSMTPHost        = ""
-	defaultMinServers      = 2
 	defaultMaxVotedTickets = 1000
 	defaultDescription     = ""
 	defaultDesignation     = ""
@@ -329,7 +328,6 @@ func loadConfig() (*config, []string, error) {
 		PublicPath:      defaultPublicPath,
 		TemplatePath:    defaultTemplatePath,
 		SMTPHost:        defaultSMTPHost,
-		MinServers:      defaultMinServers,
 		MaxVotedTickets: defaultMaxVotedTickets,
 		Description:     defaultDescription,
 		Designation:     defaultDesignation,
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -1,7 +1,8 @@
 config.toml
 dcrstakepool
-.vendor/
+vendor/
```
jholdstock: I believe this line could be removed entirely

itswisdomagain: I use the vendor folder locally within my IDE to enable me explore source code for external packages easily. Doesn't hurt to keep it, lest I or someone else mistakenly checks their vendor folder into a PR.

JoeGruffins:
I would rather see cfg.MinServers be made a constant of 2 rather than kept as a setting, as this seems to be established in the code, the readme, the sample-dcrstakepool.conf, and here https://docs.decred.org/advanced/operating-a-vsp/ Actually three but we don't have a stakepoold running on the front end yet (maybe in the future?).

itswisdomagain:
> This is a good point from Joe. I thought we already had a check in `config.go` for `cfg.MinServers > 1`, but it seems like we do not. I think that should be added before these changes are merged.

This is fine. Making the minimum number of required backend servers constant in code doesn’t sound too good.

Also, is it okay to allow `cfg.MinServers == 1` for testnet/simnet (to allow quick setup on a development machine) and enforce `cfg.MinServers > 1` for mainnet?

JoeGruffins:
It's really kinda pointless as a setting at the moment. We've taken the stance that if any of the stakepoold servers don't work, throw an error. So the real minimum is the number of servers set.

jholdstock:
My personal opinion is that we should keep the minimum as it is right now - 2 in all cases.

Having different minimum config between sim/test/main networks is a good way to enable defects to not manifest in dev or testing, and only get noticed when deployed onto mainnet.

I'm happy to be convinced otherwise on this point if others feel strongly.

config.go
```diff
@@ -582,9 +582,15 @@ func loadConfig() (*config, []string, error) {
...
-	if len(cfg.WalletHosts) < 2 {
-		str := "%s: you must specify at least 2 wallethosts"
-		err := fmt.Errorf(str, funcName)
+	// Check if deprecated minservers config option is set
+	if cfg.MinServers != 0 {
+		str := "%s: Config minservers is deprecated.  Please remove from your config file"
+		log.Warnf(str, funcName)
+	}
+
+	if len(cfg.WalletHosts) < minRequiredBackendServers {
```
JoeGruffins: There is already a check for > 0. 
https://github.com/decred/dcrstakepool/blob/0a62ba5d4a3712b1085eaeb4d10f4ee2393b3276/config.go#L546

The testnet and simnet checks won't matter at 1.  And @jholdstock still wants an argument as to why testnet should be different that mainnet.

jholdstock: I think keeping both checks is reasonable. It allows the error output to be more detailed and more useful for the reader

JoeGruffins: ok. After some soul searching, I am also ok with this change.

PR #465, opened 2019-08-09
Add links for all dcrwallet repo ref in readme

amassarwi:
- Replaced all text ref with actual link in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,19 +50,19 @@ vote on their behalf when the ticket is selected.
 
 ## 1.1.1 Release Notes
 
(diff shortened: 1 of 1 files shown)


JoeGruffins:
Other links only occur on the first use of the word/idea.  tickets, Decred, modules, etc. Is there a need for more links?

amassarwi:
Gottcha, maybe not for release note and upgrade guide sections, but it could be helpful if the ref in the 'Pre-requisites' was still a link instead of just text.

Feel free to close the PR in case you find it's unneeded

PR #466, opened 2019-08-10
maxusers: Update TODOs

### LINKED ISSUE #310: max users needs to be a config setting or project wide const
We need to get around to making this a config setting at some point.  Note that inaddition to the 10000 magic number appearing various config files for address derivation, it's also set as a package-level var in the `controller` package: https://github.com/decred/dcrstakepool/blob/f6b21eee683a50f109e367625e540b42c7ffe508/controllers/main.go#L46

_Originally posted by @chappjc in https://github.com/decred/dcrstakepool/pull/308_

JoeGruffins:
~-make max users a configurable variable~
~-default 10000~
~-stakepoold start with default 10000~
~-when starting dcrstakepool grpc commands are used to set the max on all servers and re-derive pool fees if there was a change~

~ahh, meant to do the draft thing, adding WIP to title~

This pr took a different direction mid-stream.  I now think that maxusers is an artificial ceiling and it should be removed. If my understanding is incorrect, please comment.

### DIFF
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	// TODO: sync with controllers/main.go MaxUsers
+	// TODO: Remove this.
 	numServicePaymentFeeAddresses uint32 = 10000
 )
 
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -44,6 +44,7 @@ const (
 	// MaxUsers is the maximum number of users supported by a voting service.
 	// This is an artificial limit and can be increased by adjusting the
 	// ticket/fee address indexes above 10000.
+	// TODO Remove this limitation by deriving fee addresses from an imported xpub.
 	MaxUsers = 10000
 	// agendasCacheLife is the amount of time to keep agenda data in memory.
 	agendasCacheLife = time.Hour
--- a/sample-dcrwallet.conf
+++ b/sample-dcrwallet.conf
@@ -1,6 +1,8 @@
 ; Derive 10000 addresses from the specified extended public key for use
 ; as user's payment addresses to the cold wallet/fee collecting wallet.
 ; xpub portion needs to match coldwalletextpub from dcrstakepool configuration.
+; The amount of addresses MUST be set to 10000. This limitation will be removed
+; in a future release.
 ;stakepoolcoldextkey=xpub:10000
 
 ; Fees as a percentage. 7.5 = 7.5%.  Precision of 2, 7.99 = 7.99%.


backend/stakepoold/rpc/api.proto
```diff
@@ -169,3 +169,8 @@ message GetStakeInfoResponse {
 	double ProportionMissed = 15;
 	uint32 Expired = 16;
 }
+
+message SetMaxUsersRequest {
+	uint32 MaxUsers = 1;
+}
+message Empty{}
```
JoeGruffins: @jholdstock 
What do you think about this? If we don't need a value, this can be reused, but it is different from the other code. I can change it to SetMaxUsersResponse.

sample-dcrstakepool.conf
```diff
@@ -143,3 +127,22 @@ designation=YourVSP
...
+
+; Whether to set the secure flag on cookies.  If you have SSL/TLS setup then
+; you should change this to true.
+;cookiesecure=true
+
+; Path to the root folder/directory which contains CSS/fonts/images/javascript.
+;publicpath=public
+
+; Path to the root folder/directory which contains the HTML templates.
+;templatepath=views
```
JoeGruffins: I was going to rearrange this, but maybe should do that in another pr

JoeGruffins:
I think this is finished. It will change a bit after #463 if that goes through first.

backend/stakepoold/rpc/api.proto
jholdstock: I would suggest using a dedicated `SetMaxUsersResponse`, even if it includes no fields right now. That will enable adding fields to the response in the future without incrementing the API major version number.

JoeGruffins: Hadn't even thought about that. ok.

JoeGruffins:
Will rebase after release.

JoeGruffins:
@jholdstock Let me know your thoughts:
Q. What is max users there for?
My A. To decide the number of fee addresses that should be derived.
Q. Should max users be settable?
A. Yes. As dcrstakepool is now, a pool that is approaching max users must be able to increase this to allow more users.
Q. Should it be set in one place?
A. No preference from me.  But I believe you think this could lead to bugs later on. In which case the appropriate path to take is add maxusers to the stakepoold conf and check that they are all the same on startup. If not let the operator change his/her conf.
Q. What's the future of max users?
A. Lets remove it.  I believe there was talk yesterday on setting all of the fee addresses inside dcrwallet. It shouldn't be done there.  Having a limit on users because of the feeaddresses (if that is the reason for this config setting) is silly.

JoeGruffins:
This pr won't work. stakepoolcoldextkey must also be set on wallet start-up and there's no way to query/set the value afterwards. As I think this entire concept will be changed in the end it's probably not worth making this value settable through rpc commands. As for this pr and the issue, either we should make it settable also in stakepoold conf or make it a constant for now. currently max users can't exceed 10000, even if set above, so making it a constant should not have an affect on any current pools.

JoeGruffins:
If there's no good way to check that dcrwallet.conf value, I think making this settable is a bad idea. I'm going to make maxusers a constant unless there is an objection.

JoeGruffins:
Ok. I think #310 should be closed in order to progress in the direction of  #504.

PR #473, opened 2019-08-13
controller/main.go: add agenda cache

### LINKED ISSUE #471: Cache dcrdata responses
Since #445 dcrstakepool is hitting dcrdata to lookup vote status. This response should be cached, probably for quite a long time.

JoeGruffins:
closes #471

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -16,6 +16,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/dchest/captcha"
@@ -44,6 +45,8 @@ const (
 	// This is an artificial limit and can be increased by adjusting the
 	// ticket/fee address indexes above 10000.
 	MaxUsers = 10000
+	// agendasCacheLife is the amount of time to keep agenda data in memory.
+	agendasCacheLife = time.Hour * 24
 )
 
 // MainController is the wallet RPC controller type.  Its methods include the
@@ -75,6 +78,24 @@ type MainController struct {
 	maxVotedTickets      int
 	description          string
 	designation          string
+}
+
+// agendasCache should only be accessed through MainController.agendas().
+var agendasCache agendasMux
+
+// agenda links an agenda to its status. Possible statuses are upcoming,
+// in progress, finished, failed, or locked in.
+type agenda struct {
+	Agenda chaincfg.ConsensusDeployment
+	Status string
+}
+
+// agendasMux holds the current available agendas.  Lock must be held for
+// read/writes.
+type agendasMux struct {
+	sync.Mutex
+	timer   time.Time
+	agendas *[]agenda
 }
 
 // Get the client's real IP address using the X-Real-IP header, or if that is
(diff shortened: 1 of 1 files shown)


controllers/main.go
```diff
@@ -185,6 +206,39 @@ func (controller *MainController) getNetworkName() string {
 	return controller.params.Name
 }
 
+// agendas() returns agendas and their statuses. Fetches agenda status from
```
jholdstock: Remove parentheses from function names in comments, as per [Effective Go](https://golang.org/doc/effective_go.html#commentary)

controllers/main.go
```diff
@@ -44,6 +45,8 @@ const (
 	// This is an artificial limit and can be increased by adjusting the
 	// ticket/fee address indexes above 10000.
 	MaxUsers = 10000
+	// agendasCacheLife is the amount of time to keep agenda data in memory.
+	agendasCacheLife = time.Hour * 24
```
jholdstock: 24 hours may be a little excessive, I was thinking more like 1 hour would suffice

controllers/main.go
```diff
@@ -185,6 +206,39 @@ func (controller *MainController) getNetworkName() string {
...
+	defer agendasCache.Unlock()
+	agendasCache.Lock()
+	now := time.Now()
+	if agendasCache.timer.After(now) {
+		return *agendasCache.agendas
+	}
+	url := fmt.Sprintf("https://%s.dcrdata.org/api/agendas", controller.getNetworkName())
+	agendaInfos, err := dcrDataAgendas(url)
+	if err != nil {
+		log.Warnf("unable to retrieve data from %v: %v", url, err)
```
jholdstock: With the way this is currently written, a failed call to dcrdata will result in the cached data being overwritten with an empty array, and then no further attempts to contact dcrdata will occur until the timeout has elapsed. If an error occurs we should probably return the old cached data after logging the warning. That way, next time `agendas()` is invoked, the call to dcrdata will be retried.

JoeGruffins:
@jholdstock Thanks for the review! I believe I covered all of your comments.

controllers/main.go
```diff
@@ -185,6 +207,45 @@ func (controller *MainController) getNetworkName() string {
...
+	agendasCache.timer = now.Add(agendasCacheLife)
+	url := fmt.Sprintf("https://%s.dcrdata.org/api/agendas", controller.getNetworkName())
+	agendaInfos, err := dcrDataAgendas(url)
+	if err != nil {
+		// Ensure the next call tries to fetch statuses again.
+		agendasCache.timer = time.Time{}
+		log.Warnf("unable to retrieve data from %v: %v", url, err)
+		// If we have initialized agendas, return that.
+		if agendasCache.agendas != nil {
+			return *agendasCache.agendas
```
jholdstock: If an error occurs we should always return here, regardless of what is cached. All of the code beyond this point assumes a successful result has been returned from dcrdata, but it hasnt.

JoeGruffins: I wrote it so that an error here will perform exactly like it does now. We still have all the  agenda info except for statuses. If this fails it will use cache if available, and not show statuses if not.

JoeGruffins: If you could try changing the url to something silly, you'll see it log an error and show the agendas as they are now.

jholdstock: Tested thoroughly, you are correct. Its different to how I would have done it, but it works and that is what matters :+1:  Approving

PR #476, opened 2019-08-13
add tmux test harness

jholdstock:
See changes in `README.MD` for further info

### DIFF
--- a/harness.sh
+++ b/harness.sh
@@ -57,6 +57,8 @@ EOF
 echo "Starting dcrd master node"
 tmux send-keys "dcrd -C ${NODES_ROOT}/master/dcrd.conf" C-m 
 
+sleep 3 # Give dcrd time to start
+
 #################################################
 # Setup multiple back-ends.
 #################################################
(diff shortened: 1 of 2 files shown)


jholdstock:
This will need to be rebased if #470 goes in first. The dcrwallet configs will need to be removed from dcrstakepool cfg file

harness.sh
```diff
@@ -0,0 +1,179 @@
...
+    #################################################
+    # stakepoold
+    #################################################
+
+    echo ""
+    echo "Writing config for stakepoold-${i}"
+    mkdir -p "${NODES_ROOT}/stakepoold-${i}"
+    cat > "${NODES_ROOT}/stakepoold-${i}/stakepoold.conf" <<EOF
+dcrdhost=${DCRD_RPC_LISTEN}
+dcrdcert=${WALLET_RPC_CERT}
```
JoeGruffins: I think this should be the dcrd cert.

jholdstock: Well that was dumb... fixed now

PR #481, opened 2019-08-16
stakepooldclient: more informative errors

### LINKED ISSUE #474: Better error messages from stakepoold client
Rather than simply an index in our error output like `stakepoold ID 1 returned an error` we can provide more useful information, such as `stakepoold at 123.456.78.9 returned an error`

jholdstock:
Closes #474


stakepooldclient/stakepooldclient.go
```diff
@@ -45,8 +48,10 @@ type StakepooldManager struct {
 	cachedStakeInfoMutex sync.Mutex
 }
 
+// ConnectStakepooldGRPC established a gRPC connection with all provided
```
JoeGruffins: establishes

jholdstock: sorted

PR #489, opened 2019-08-18
Changed NewMainController signature to accept params struct hence red…

### LINKED ISSUE #399: NewMainController - pass a Controller Config
NewMainController takes a slew of parameters.  It may make it Config struct and pass that instead.

UferePease:
Changed NewMainController signature to accept params struct hence reducing the number of params passed to it. Refactored MainController struct to inherit from the newly created MainControllerParams struct

Closes #399


JoeGruffins:
Hello!

your comment message is rather long. For future reference https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

controllers/main.go
```diff
@@ -49,34 +49,41 @@ const (
 	agendasCacheLife = time.Hour
 )
 
+// Config struct to be passed as argument into NewMainController func
+// Reduces the number of marameters passed to NewMainController()
```
JoeGruffins: "parameters"

JoeGruffins:
My personal opinion is that this doesn't really simplify things.  But it is working fine.

UferePease:
> Hello!
> 
> your commit message is rather long. For future reference https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

Thanks

controllers/main.go
```diff
@@ -49,34 +49,41 @@ const (
...
-	realIPHeader         string
 	captchaHandler       *CaptchaHandler
 	emailSender          email.Sender
 	voteVersion          uint32
 	votingXpub           *hdkeychain.ExtendedKey
-	maxVotedTickets      int
-	description          string
-	designation          string
+
+	mainControllerParams MainControllerParams
```
jholdstock: Lets rename this field `params`.

Calling it `mainControllerParams` means we need to write code like `mainController.mainControllerParams` when we could be writing simply `mainController.params`. This is easier read and write whilst being just as descriptive.

The [naming section of Effective go](https://golang.org/doc/effective_go.html#names) provides useful guidelines on this topic.

server.go
```diff
@@ -98,12 +98,25 @@ func runMain() error {
...
+		Designation:     cfg.Designation,
+	}
+
 	controller, err := controllers.NewMainController(activeNetParams.Params,
-		cfg.AdminIPs, cfg.AdminUserIDs, cfg.APISecret, APIVersionsSupported,
-		cfg.BaseURL, cfg.ClosePool, cfg.ClosePoolMsg, coldWalletFeeKey,
-		stakepooldConnMan, cfg.PoolFees, cfg.PoolEmail, cfg.PoolLink,
-		sender, cfg.RealIPHeader, votingWalletVoteKey,
-		cfg.MaxVotedTickets, cfg.Description, cfg.Designation)
+		mainControlParam, APIVersionsSupported, coldWalletFeeKey,
```
jholdstock: why are `APIVersionsSupported`, `coldWalletFeeKey` (and others) being left out of the new config struct?

controllers/main.go
JoeGruffins: Or just export everything in the original struct and do the initialization in main package? That was one of my suggestions in the issue, but I dunno if that's what @dajohi had in mind.

server.go
UferePease: They have been included in the new commit.

controllers/main.go
UferePease: There is already a `params` field in MainController struct. I have pushed a new commit.

jholdstock:
> My personal opinion is that this doesn't really simplify things. But it is working fine.

The main benefit of making this change is transforming this section:

Screenshot from 2019-08-20 12-05-32

Changing this awful one liner and splitting it into a line per value makes the code significantly easier to read, edit and review.

controllers/main.go
```diff
@@ -49,34 +49,41 @@ const (
...
-	params               *chaincfg.Params
-	realIPHeader         string
-	captchaHandler       *CaptchaHandler
-	emailSender          email.Sender
-	voteVersion          uint32
-	votingXpub           *hdkeychain.ExtendedKey
-	maxVotedTickets      int
-	description          string
-	designation          string
+	enableStakepoold bool
```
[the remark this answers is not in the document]
UferePease: Okay. Done

JoeGruffins:
> Changing this awful one liner and splitting it into a line per value makes the code significantly easier to read, edit and review.

Ok, I agree there. But why not use the original struct? Or embed the new struct. Then there's no reason for MainControllerParams.Params

JoeGruffins:
Ok! I think it'll be good after a rebase. Perhaps you merged master when what you needed was a rebase? You can move comments around and delete with an interactive rebase.

UferePease:
I must have mixed things up here with my rebase, I should open a fresh PR

UferePease:
> I must have mixed things up here with my rebase, I should open a fresh PR

The new PR is here: #492 
So sorry

Issue #269, opened 2018-09-03
Stakepools design overhaul: GDPR, naming consensus and other items


linnutee:
This issue is a base point for informing and coordinating the release of a design update for stakepools theme. The update mainly focuses on improving the user experience for staking, as well touches point with some of the stakepool administrative items and the recent case around terminology. https://github.com/decred/dcrdesign/issues/23

Design work, largely created by @MariaPleshkova can be previewed here: https://www.figma.com/file/dzjw34JZ40sxEJVqhekmhZBP/Decred
Currently these views are being coded into front-end by @rmykolas, and lastly will be tied to the stakepool dynamics by @karamble 

Once the process is further, plan is to give a another notice well ahead before the release to the stakepool hosts. 

**Naming transitions**
There appears to be a rough consensus to use Voting Service Provider (VSP) instead of stakepool. As this is a larger change, requesting feedback from stakepool hosts. Please respond with your opinions and considerations or suggestions. E.g. does this create any serious conflicts with your domains?

Touching point on related design improvements here. Taking into account recognisability of individual stakepools. Current state around distinguishing stakepools by either domain, self-picked name and ID's is inconsistent. The current release only has standard DCR logo – so there's little differentiation options between the pools. However in the design update, we've created a derivative logo lockup to highlight the network id of each stakepools. 

Obviously some of stakepools have put further effort down into naming and branding. This is obviously an individual interest already. The intent of the update is rather to provide all stakepool hosts  an improved and more even starting point. As well consider potential of building the communications around the stakepool ID rather than having several competing names. This paints a more professional and established picture for new users picking between stakepools.

<img width="1268" alt="screen shot 2018-09-03 at 17 23 00" src="https://user-images.githubusercontent.com/17774057/44992176-2cbc5100-af9f-11e8-837f-bd41ca332db8.png">

**GDPR:** there is a standing issue needing attention: https://github.com/decred/dcrstakepool/issues/259 
Requesting help from copywriters to get involved and put together a generic privacy policy terms. The new theme should not be launched without it. For education/reference: https://twitter.com/smashingmag/status/999189181162250241

linnutee:
Any update on how the design design implementation is going @karamble or has someone else picked this up? Asking so the release of this design update would be somewhat coordinated w/ Decrediton and dcrweb, thanks.

https://github.com/decred/dcrdesign/issues/23

jholdstock:
I've spoken with @karamble and @linnutee, I am going to take this on.

[...]

PR #491, opened 2019-08-19
input new styling

### LINKED ISSUE #385: [Design] Styling for input validation
_Copying from @MariaPleshkova in #339 and further information available in [the new design](https://github.com/decred/dcrstakepool/issues/269)_

When there are problems with data entered (e.g. incorrect login/email, field is empty,etc.), notification should be like this:

amassarwi:
Still in progress, would like to get some feedback regarding the approach before changing all other inputs/validations 

Open tasks:
- [x] Add icon
- [x] Change/update all other "flash" messages listed in @jholdstock comment - [here](https://github.com/decred/dcrstakepool/issues/269#issuecomment-482801298)

* Closes #385


controllers/main.go
```diff
@@ -1595,13 +1595,14 @@ func (controller *MainController) LoginPost(c web.C, r *http.Request) (string, i
 	user, err := helpers.Login(dbMap, email, password)
 	if err != nil {
 		log.Infof(email+" login failed %v, %v", err, remoteIP)
-		session.AddFlash("Invalid Email or Password", "loginError")
```
amassarwi: So as you can see I'm using `c.Env["IsInvalidLogin"]` is there a reason why error messages were handled as session flash messages ? with `c.Env["IsInvalidLogin"]` I was able to toggle the error message div (show/hide) but also add/remove class from the input to change the styling when error occurs.
Would like to get your opinion before moving forward with my implementation.

amassarwi:
Added icon as well:
image

public/css/styles.css
```diff
@@ -8772,6 +8782,11 @@ https://flickity.metafizzy.co
   cursor: pointer;
 }
 
+.icon-warning {
+  width: 14px;
+  height: 14px;
+  background-image: url(/assets/images/warning-sign.svg); }
```
amassarwi: I used the existing warning .svg - however, the one shown in #385 design/screenshot looks different - do we have it somewhere as .svg file as well ?

controllers/main.go
JoeGruffins: Perhaps to allow the error string to be dynamic?  And also it's gorilla, so it's cool.  Something about good for redirects. I would think you can still do the add/remove stuff with the flash message, but I don't have a strong opinion on this. It looks fine to me.

public/css/styles.css
amassarwi: @jholdstock - maybe you can navigate here ?

controllers/main.go
amassarwi: Agree regarding the flash messages - perhaps I could use the value of the flash message instead of the new boolean to achieve the class toggle as well - will give a try

views/auth/login.html
```diff
@@ -10,16 +10,22 @@
...
+        <div class="mb-4 mx-auto w-75">
+          <div class="row justify-content-center flex-nowrap mx-0">
+            <input type="email" name="email" class="form-control {{if .FlashError}} err-form-control {{end}}" placeholder="Email" required autofocus>
+            {{if .FlashError}} 
+              <div class="icon-warning"></div>
+            {{end}}
+          </div>
+          {{if .FlashError}}
+            <div class="err-message mt-1 ml-1 text-left">
+                {{range .FlashError}}
```
amassarwi: Switched back to the FlashError approach (dynamic messages) and was able to toggle the styling classes with if statement as well.

Below you find the second possible error message in the login page `You must validate your email address`
image
`
image

public/css/styles.css
amassarwi: Added new style icon as well

amassarwi:
- added new icon provided in style
- I have styled the flash messages with new style in two pages where the form is not rested to captcha screen (login and passwordupdate, is there more which behave the same?)

**This should be the scope of this PR, unless there are more forms which behave like the one already updated**

controllers/main.go
```diff
@@ -1277,10 +1277,9 @@ func (controller *MainController) PasswordResetPost(c web.C, r *http.Request) (s
 	if !controller.IsCaptchaDone(c) {
 		session.AddFlash("You must complete the captcha.", "passwordresetError")
 		return controller.PasswordReset(c, r)
-	} else {
-		session.Values["CaptchaDone"] = false
-		c.Env["CaptchaDone"] = false
 	}
+	session.Values["CaptchaDone"] = false
+	c.Env["CaptchaDone"] = false
```
JoeGruffins: Was this wrong? It seems unrelated to the issue.

amassarwi: Not related directly but saw it while working on the task, there was a redundant else statement, anyway these two lines will run if the condition is false, so I fixed it.

jholdstock:
Disabling native error messages with me is fine, just two things to consider:

- Needs to work on all browsers
- Needs to degrade nicely without javascript

JoeGruffins:
Last time I checked on windows, the site didn't work at all.  So, I wouldn't worry about it.

public/images/warning-sign.svg
```diff
@@ -1,4 +1,3 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" fill="none" viewBox="0 0 14 14">
-    <path stroke="#FD704A" stroke-width=".98439" d="M7 .492A6.506 6.506 0 0 1 13.508 7 6.507 6.507 0 1 1 7 .492z"/>
-    <path fill="#FD704A" d="M5.706 10.084a1.288 1.288 0 0 1 2.576 0 1.288 1.288 0 1 1-2.576 0zm.405-5.76l-.063-1.712h1.89l-.063 1.712-.263 3.629H6.37l-.259-3.63z"/>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 450 450" width="20" height="20">
+    <path d="M225.24,24.86c-110.46,0-200,89.55-200,200s89.54,200,200,200,200-89.55,200-200S335.69,24.86,225.24,24.86Z" fill="#ed6d47"></path><rect fill="none"></rect><path d="M188.2,313c0-21.92,15.7-36.84,36.8-36.84s36.8,14.92,36.8,36.84-15.7,37.47-36.8,37.47S188.2,335,188.2,313Zm11.56-164.61L198,99.49H252l-1.8,48.94-7.5,103.67H207.26Z" fill="#c4cbd2"></path>
```
jholdstock: > fill="#c4cbd2"

This results in a gray `!` over an orange background. If we update to 

> fill="#fff"

We will have a white `!`, which I think looks much better. Comparison below

Screenshot from 2019-09-23 12-58-20

amassarwi: @jholdstock Thanks for the feedback, regarding icon coloring it looks much better with white `!`, updating it.

JoeGruffins:
@amassarwi needs rebase again

JoeGruffins:
@amassarwi 
That replace I told you to put in the go.mod is happening twice now it seems.  Sorry for all the rebase trouble.  Someone else got that change in before us.

amassarwi:
@JoeGruffins All good, reverted

JoeGruffins:
I may be doing something incredibly stupid, but this doesn't seem to be working for me atm.

JoeGruffins:
I noticed something I hadn't before, the very right of the exclamation circle is cut off testing on mobile using chrome and duck duck go

<details close>
<summary>pic</summary>

Screenshot_20191002-074352

</details>

Also, I guess it wasn't possible to hide the message once a user starts to type again?

JoeGruffins:
Sorry to bring up a problem without having a good answer. I don't know.  If it's impossible that's ok.  It's not that important.

jholdstock:
A little bit of extra javascript to remove the warning after typing should be fine imo. Obviously that wont work when js is disabled, but thats acceptable.

jholdstock:
The fading code doesn't seem to work very well for me in chrome. Often when I submit the login form with an incorrect password, the error message appears as expected, but it fades out immediately without me taking any further actions. It also fades out if I click anywhere on the screen which seems a bit strange.

Now that I've seen this behavior, I actually think I prefer it without the automatic fade out. How strongly do you feel about keeping this joe/amassarwi?

amassarwi:
@jholdstock Couldn't really reproduce the behavior you described above.
In my case the error fades out only when I start typing or when I paste new value into input field and it works pretty well.

It's nicer ux indeed to hide the error, but can't really reproduce the strange behavior to able to fix it.

jholdstock:
ffs... I was working with a keyboard I don't normally use. I forgot it has a sticky `CTRL` key, and that was causing the error message to fade. Testing with a proper keyboard this feature works nicely. 

I'm still not sure about the fade on clicking. I guess this is due to the text input losing focus. Is there anything we can do about that?

jholdstock:
Figured it out - its my browser auto-filling my email address when the form loads. That triggers the onchange handler.

I guess we can accept this as-is, because we cannot do anything about browser auto-filling.

@amassarwi will you proceed with the rest of the forms in this PR or create a new one?

amassarwi:
I have disabled the auto complete for the form using the html attribute mentioned in the previous comment, I would suggest we keep it unless someone has good arguments why shouldn't we, what do you think ?

amassarwi:
If the auto-complete="off" is not working in all browsers, then I may just revert it before we merge.

PR #493, opened 2019-08-20
Check coldwalletextpub

### LINKED ISSUE #450: Check that all coldwalletextpub are the same accross all config files
The value set here
https://github.com/decred/dcrstakepool/blob/cce2438c018237ee12d13fef64f96c0ee88e41e1/sample-dcrstakepool.conf#L63
and the values set here
https://github.com/decred/dcrstakepool/blob/cce2438c018237ee12d13fef64f96c0ee88e41e1/sample-stakepoold.conf#L5
should all be the same.  Setting them as different keys creates problems that are difficult to track down.
The best answer would be to set them all in one place, but the easy answer is to just check that they are all the same.

UferePease:
Check that coldwallettextpub in stakepoold.conf matched that in dcrstakepool.conf.
Should fix #450


JoeGruffins:
coldwalletextpub is on a different physical machine. Also, it will already be stored in memory. You shouldn't need to read the config again.

dajohi:
This code is not the suggestion in #450

backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -53,6 +53,7 @@ type AppContext struct {
 	WalletConnection       *rpcclient.Client
 	WinningTicketsChan     chan WinningTicketsForBlock
 	Testing                bool // enabled only for testing
+	ColdWalletExtPub       string
```
itswisdomagain: I'll prefer this be moved to after line 42 and similarly assigned after line 204 in `backend/stakepoold/server.go`. Both are simple copies of config values from the `cfg` object to the `AppContext` instance; keeping them close together makes it slightly easier to read/understand imo.

backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -422,6 +423,12 @@ func (ctx *AppContext) GetStakeInfo() (*wallettypes.GetStakeInfoResult, error) {
 	return response, nil
 }
 
+// GetColdWalletExtPub performs the rpc command GetColdWalletExtPub.
+func (ctx *AppContext) GetColdWalletExtPub() string {
+	response := ctx.ColdWalletExtPub
+	return response
```
itswisdomagain: Since `ctx.ColdWalletExtPub` is exported, is there really a need for `ctx.GetColdWalletExtPub()`?
It does fall in line with the existing pattern for implementing rpc commands. If keeping this function, can just `return ctx.ColdWalletExtPub` instead of creating and assigning `response` variable.

backend/stakepoold/rpc/rpcserver/server.go
```diff
@@ -279,3 +279,10 @@ func (s *stakepooldServer) GetStakeInfo(ctx context.Context, req *pb.GetStakeInf
 		Expired:          response.Expired,
 	}, nil
 }
+
+func (s *stakepooldServer) GetColdWalletExtPub(ctx context.Context, req *pb.GetColdWalletExtPubRequest) (*pb.GetColdWalletExtPubResponse, error) {
+	response := s.appContext.GetColdWalletExtPub()
+	return &pb.GetColdWalletExtPubResponse{
+		ColdWalletExtPub: response,
```
itswisdomagain: Can inline this: `ColdWalletExtPub: s.appContext.GetColdWalletExtPub(),`
or as suggested in an earlier review comment `ColdWalletExtPub: s.appContext.ColdWalletExtPub,`

config.go
```diff
@@ -22,41 +22,44 @@ import (
...
-	defaultPublicPath      = "public"
-	defaultTemplatePath    = "views"
-	defaultSMTPHost        = ""
-	defaultMaxVotedTickets = 1000
-	defaultDescription     = ""
-	defaultDesignation     = ""
+	defaultBaseURL                  = "http://127.0.0.1:8000"
+	defaultClosePoolMsg             = "The voting service is temporarily closed to new signups."
+	defaultConfigFilename           = "dcrstakepool.conf"
+	defaultStakepooldConfigFilename = "stakepoold.conf"
```
itswisdomagain: Don't think this is needed anymore. Remove this and related additions to this file. Remember to `go fmt ./...` afterwards.

UferePease: changes applied.

JoeGruffins:
Again, I think you merged master when you should of rebased. This can be fixed with an interactive rebase and forced push.

itswisdomagain:
Completed in #526.

Issue #475, opened 2019-08-13
Handle stakepoold/dcrwallet/dcrd disconnects


jholdstock:
Reported by @JoeGruffins:

>But if dcrwallet is not running, it seems any command (CreateMultisig at least) will hang without error. Dcrstakepool will not start up if a wallet is not online, but if one goes offline after starting dcrstakepool, stakepoold starts trying to reconnect, and it seems, maybe all commands, will wait for a reconnect indefinitely, and then fire. The user will see a 504 gateway timeout error, and the operator will get no notice. Maybe this is a problem for another issue. I do not expect this pr is the cause, but if you can also check for stakepoold's trying to reconnect to a wallet state, I think that would be good...

jholdstock:
#476 has been raised to facilitate this. A proper test harness removes the effort required for repeatedly creating/tearing down full VSP configurations.

jholdstock:
Related #166 and #70

JoeGruffins:
As for the dcrd and dcrwallet disconnects, I think our problem lies here
https://github.com/decred/dcrd/blob/8497b9843bcb4191ec1f4a235da178660fb9ad70/rpcclient/infrastructure.go#L655
We are waiting for a reconnect and then all commands are sent.
Some possible solutions:
- Add a timeout to grpc commands. If they take to long throw a timeout error and discard whatever results may come back at a later time.
- Ping dcrwallet/dcrd before sending commands. This reinstates the dcrwallet link we just broke though.
- Turn off the auto reconnect feature and rewrite our own auto-reconnect that allows us to take the status of the connection.
@jholdstock I think the third option is best, or can you think of any more.

jrick:
The third option is what I had done before in dcrwallet (while it was still using rpcclient -- as of master it no longer does) and I recommend that approach.

jholdstock:
I have done a whole bunch of manual testing using the framework below. I think we want to resolve all of these before cutting a release.

### Disconnect dcrd

|Test|Pass|Notes|
|----|----|-----|
|Check disconnection in logs|:heavy_check_mark:||
|Check `/status` for DaemonConnected == false|:heavy_check_mark:||
|All pages should still load|:heavy_check_mark:||
|Can still log-in and register|:heavy_check_mark:||
|Can still update voting preferences, change password|:heavy_check_mark:||
|Error shown when trying to connect decrediton.|:heavy_check_mark:||


### Restart dcrd

|Test|Pass|Notes|
|----|----|-----|
[...]

PR #494, opened 2019-08-20
rpcserver: Add walletConnected method

JoeGruffins:
Check whether the wallet is connected before sending rpc commands.
part of #475 

Perhaps this is the easiest solution. It's solving the original problem that I described in that issue. Simply check that the wallet is connected before continuing to send rpc commands.  Another solution I suggested was to write our own reconnect function, but maybe that's not necessary to solve this problem.

### DIFF
--- a/backend/stakepoold/rpc/rpcserver/server.go
+++ b/backend/stakepoold/rpc/rpcserver/server.go
@@ -13,6 +13,7 @@
 package rpcserver
 
 import (
+	"errors"
 	"time"
 
 	"golang.org/x/net/context"
@@ -37,6 +38,12 @@ const (
 	semverMajor        = 7
 	semverMinor        = 0
 	semverPatch        = 0
+)
+
+var (
+	// ErrWalletNotConnected is an error to describe the condition where
+	// dcrwallet is not connected through dcrcleint.
+	ErrWalletNotConnected = errors.New("wallet is disconnected")
 )
 
 // versionServer provides RPC clients with the ability to query the RPC server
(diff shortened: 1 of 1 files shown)


backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -172,6 +172,17 @@ func (ctx *AppContext) EvaluateStakePoolTicket(tx *wire.MsgTx, blockHeight int32
 	return true, nil
 }
 
+// walletConnected logs and returns an error if the connection to dcrwallet is
+// broken.
+func (ctx *AppContext) walletConnected() error {
```
jholdstock: Lets return a bool here rather than an error

JoeGruffins:
Moved to server.go and changed to return a bool.  Also changed the comment to explain that this is a temporary fix.

jholdstock:
### Kill one instance of dcrwallet

|Test|Pass|Notes|
|----|----|-----|
|Check disconnection in logs|:heavy_check_mark:||
|All pages should load|:heavy_check_mark:||
|`/status` shows the disconnected wallet|:heavy_check_mark:||
|Change voting preferences without error|::heavy_check_mark:||
|Changing password works|:heavy_check_mark:||
|Login`/register` works|:heavy_check_mark:||
|Error shown when trying to connect decrediton.|:heavy_check_mark:||

### Reconnect dcrwallet:

|Test|Pass|Notes|
|----|----|-----|
|Check for reconnection in logs|:heavy_check_mark:|Reconnection takes up to 1 minute|
|Connect decrediton works|:x:|Error "system error - unable to process wallet commands" (but it does work after restarting dcrstakepool|
|All pages load|:heavy_check_mark:||
|`/status` shows wallet status|:heavy_check_mark:||
|Change voting preferences without error|:heavy_check_mark:||

jholdstock:
#498

PR #510, opened 2019-08-29
Autoreconnect

JoeGruffins:
Turn off rpcclient's autoreconnect and use our own.  This is done to remove the checks before ever wallet rpc call.  Checks for a disconnect every so often and starts trying to reconnect.

### DIFF
--- a/backend/stakepoold/rpc/rpcserver/client.go
+++ b/backend/stakepoold/rpc/rpcserver/client.go
@@ -1,0 +1,167 @@
+package rpcserver
+
+import (
+	"sync"
+	"time"
+
+	"github.com/decred/dcrd/rpcclient/v3"
+)
+
+const (
+	// connectionRetryInterval is the amount of time to wait in between
+	// retries when automatically reconnecting to an RPC server.
+	connectionRetryInterval = time.Second * 5
+	// disconnectCheckInterval is the amount of time to wait between
+	// checks for a disconnection.
+	disconnectCheckInterval = time.Second * 10
+)
+
+// Client holds the information related to an rpcclient and handles access to
+// that client through a mutex.
+//
+// It should be noted that this is a temporary fix to the problem that rpcclient
+// does not return an error when autoreconnect is turned on but the client is
+// disconnected. The permanent solution is to change the behaviour of rpccleint.
+// TODO: Remove this file.
+type Client struct {
+	// client is protected by a mutex that must be held for reads/writes.
+	client *rpcclient.Client
+	mux    sync.RWMutex
+
+	cfg          *rpcclient.ConnConfig
+	ntfnHandlers *rpcclient.NotificationHandlers
+	wg           sync.WaitGroup
+	stop         chan struct{}
+
+	connected    chan struct{}
+	connectedMux sync.Mutex
+}
+
+// Connected returns a receiving copy of the current connected channel. If
+// disconnected and the channel is not yet blocking, creates a new channel that
+// will be closed on a successful reconnect.
+func (c *Client) Connected() <-chan struct{} {
+	c.connectedMux.Lock()
+	defer c.connectedMux.Unlock()
+	if c.RPCClient().Disconnected() {
+		// Start blocking on connected chan if not already.
+		select {
+		case <-c.connected:
+			c.connected = make(chan struct{})
+		default:
+		}
+	}
+	return c.connected
+}
+
+// IsConnected checks and returns whethere the client is currently connected.
+func (c *Client) IsConnected() bool {
+	select {
+	case <-c.Connected():
+		return true
+	default:
+		return false
+	}
+}
+
+// RPCClient allows access to the underlying rpcclient by providing a copy of
+// its address.
+func (c *Client) RPCClient() *rpcclient.Client {
+	c.mux.RLock()
+	defer c.mux.RUnlock()
+	return c.client
+}
+
+// New creates a new Client and starts the automatic reconnection handler.
+// Returns an error if unable to construct a new rpcclient.
+func NewClient(cfg *rpcclient.ConnConfig, ntfnHandlers *rpcclient.NotificationHandlers) (*Client, error) {
+	client, err := rpcclient.New(cfg, ntfnHandlers)
+	if err != nil {
+		return nil, err
+	}
+	c := &Client{
+		client:       client,
+		cfg:          cfg,
+		ntfnHandlers: ntfnHandlers,
+		stop:         make(chan struct{}),
+		connected:    make(chan struct{}),
+	}
+	// A closed connected channel indcates successfully connected.
+	close(c.connected)
+	c.wg.Add(1)
+	go c.autoReconnect()
+	return c, nil
+}
+
+// Stop stops automatic reconnections.
+func (c *Client) Stop() {
+	c.stop <- struct{}{}
+	// Wait for autoReconnect to stop.
+	c.wg.Wait()
+	// Stop blocking on connected if blocking, as we will never reconnect again.
+	if !c.IsConnected() {
+		close(c.connected)
+	}
+}
+
+// autoReconnect waits for a disconnect or stop. On disconnect it attempts to
+// reconnect to the client every connectionRetryInterval.
+//
+// This function must be run as a goroutine.
+func (c *Client) autoReconnect() {
+out:
+	for {
+		select {
+		case <-c.stop:
+			break out
+		case <-time.After(disconnectCheckInterval):
+			if c.IsConnected() {
+				continue out
+			}
+			// Client is disconnected. Try to reconnect.
+		}
+
+	reconnect:
+		for {
+			select {
+			case <-c.stop:
+				break out
+			default:
+			}
+
+			// Start a new client.
+			client, err := rpcclient.New(c.cfg, c.ntfnHandlers)
+
+			if err != nil {
+				log.Warnf("Failed to connect to %s: %v",
+					c.cfg.Host, err)
+
+				log.Infof("Retrying connection to %s in "+
+					"%s", c.cfg.Host, connectionRetryInterval)
+				time.Sleep(connectionRetryInterval)
+				continue reconnect
+			}
+
+			// Properly shutdown old client.
+			c.mux.Lock()
+			c.client.Shutdown()
+			c.client.WaitForShutdown()
+			// Switch the new client with the old, shutdown one.
+			c.client = client
+			c.mux.Unlock()
+
+			// Close the connected channel so that all waiting
+			// processes can continue.
+			close(c.connected)
+
+			log.Infof("Reestablished connection to RPC server %s",
+				c.cfg.Host)
+
+			// Break out of the reconnect loop back to wait for
+			// disconnect again.
+			break
+		}
+	}
+	c.wg.Done()
+	log.Tracef("RPC client reconnect handler done for %s", c.cfg.Host)
+}
--- a/backend/stakepoold/ntfnhandlers.go
+++ b/backend/stakepoold/ntfnhandlers.go
@@ -30,6 +30,8 @@ func getNodeNtfnHandlers(ctx *rpcserver.AppContext) *rpcclient.NotificationHandl
 				BlockHeight: blockHeight,
 				SmTickets:   ticketsFixed,
 			}
+			// Wait for a wallet connection if not connected.
+			<-ctx.WalletConnection.Connected()
 			ctx.SpentmissedTicketsChan <- smt
 		},
 		OnWinningTickets: func(blockHash *chainhash.Hash, blockHeight int64, winningTickets []*chainhash.Hash) {
--- a/backend/stakepoold/rpc/rpcserver/context.go
+++ b/backend/stakepoold/rpc/rpcserver/context.go
@@ -50,7 +50,7 @@ type AppContext struct {
 	SpentmissedTicketsChan chan SpentMissedTicketsForBlock
 	UserData               *userdata.UserData
 	VotingConfig           *VotingConfig
-	WalletConnection       *rpcclient.Client
+	WalletConnection       *Client
 	WinningTicketsChan     chan WinningTicketsForBlock
 	Testing                bool // enabled only for testing
 }
@@ -197,7 +197,7 @@ func (ctx *AppContext) getticket(wg *sync.WaitGroup, nt *ticketMetadata) {
 	// Ask wallet to look up vote transaction to see if it belongs to us
 	log.Debugf("calling GetTransaction for %v ticket %v",
 		strings.ToLower(nt.ticketType), nt.ticket)
-	res, err := ctx.WalletConnection.GetTransaction(nt.ticket)
+	res, err := ctx.WalletConnection.RPCClient().GetTransaction(nt.ticket)
 	nt.getDuration = time.Since(start)
 	if err != nil {
 		// suppress "No information for transaction ..." errors
@@ -283,13 +283,13 @@ func (ctx *AppContext) UpdateTicketDataFromMySQL() error {
 // the new user has registered.
 func (ctx *AppContext) ImportNewScript(script []byte) (int64, error) {
 
-	err := ctx.WalletConnection.ImportScriptRescanFrom(script, false, 0)
+	err := ctx.WalletConnection.RPCClient().ImportScriptRescanFrom(script, false, 0)
 	if err != nil {
 		log.Errorf("ImportNewScript: ImportScriptRescanFrom rpc failed: %v", err)
 		return -1, err
 	}
 
-	_, bestBlockHeight, err := ctx.WalletConnection.GetBestBlock()
+	_, bestBlockHeight, err := ctx.WalletConnection.RPCClient().GetBestBlock()
 	if err != nil {
 		log.Errorf("ImportNewScript: GetBestBlock rpc failed: %v", err)
 		return -1, err
@@ -306,7 +306,7 @@ func (ctx *AppContext) ImportMissingScripts(scripts [][]byte, rescanHeight int) 
 	// Import n-1 scripts without a rescan.
 	allButOne := scripts[:len(scripts)-1]
 	for _, script := range allButOne {
-		err := ctx.WalletConnection.ImportScriptRescanFrom(script, false, 0)
+		err := ctx.WalletConnection.RPCClient().ImportScriptRescanFrom(script, false, 0)
 		if err != nil {
 			log.Errorf("ImportMissingScripts: ImportScript rpc failed: %v", err)
 			return err
@@ -315,7 +315,7 @@ func (ctx *AppContext) ImportMissingScripts(scripts [][]byte, rescanHeight int) 
 
 	// Import the last script and trigger a rescan
 	lastOne := scripts[len(scripts)-1]
-	err := ctx.WalletConnection.ImportScriptRescanFrom(lastOne, true, rescanHeight)
+	err := ctx.WalletConnection.RPCClient().ImportScriptRescanFrom(lastOne, true, rescanHeight)
 	if err != nil {
 		log.Errorf("ImportMissingScripts: ImportScriptRescanFrom rpc failed: %v", err)
 		return err
@@ -335,13 +335,13 @@ func (ctx *AppContext) AddMissingTicket(ticketHash []byte) error {
 		return err
 	}
 
-	tx, err := ctx.WalletConnection.GetRawTransaction(hash)
+	tx, err := ctx.WalletConnection.RPCClient().GetRawTransaction(hash)
 	if err != nil {
 		log.Errorf("AddMissingTicket: GetRawTransaction rpc failed: %v", err)
 		return err
 	}
 
-	err = ctx.WalletConnection.AddTicket(tx)
+	err = ctx.WalletConnection.RPCClient().AddTicket(tx)
 	if err != nil {
 		log.Errorf("AddMissingTicket: AddTicket rpc failed: %v", err)
 		return err
@@ -351,7 +351,7 @@ func (ctx *AppContext) AddMissingTicket(ticketHash []byte) error {
 }
 
 func (ctx *AppContext) ListScripts() ([][]byte, error) {
-	scripts, err := ctx.WalletConnection.ListScripts()
+	scripts, err := ctx.WalletConnection.RPCClient().ListScripts()
 	if err != nil {
 		log.Errorf("ListScripts: ListScripts rpc failed: %v", err)
 		return nil, err
@@ -374,7 +374,7 @@ func (ctx *AppContext) CreateMultisig(addresses []string) (*wallettypes.CreateMu
 		decodedAddresses[i] = decodedAddress
 	}
 
-	result, err := ctx.WalletConnection.CreateMultisig(1, decodedAddresses)
+	result, err := ctx.WalletConnection.RPCClient().CreateMultisig(1, decodedAddresses)
 	if err != nil {
 		log.Errorf("CreateMultisig: CreateMultisig rpc failed: %v", err)
 		return nil, err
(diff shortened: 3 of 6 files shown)


jholdstock:
Love it! Lets include this in the upcoming release rather than including our temporary solution.

JoeGruffins:
> Love it! Lets include this in the upcoming release rather than including our temporary solution.

Thanks! I haven't tested much yet tho.

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,127 @@
...
+				log.Infof("Failed to connect to %s: %v",
+					c.cfg.Host, err)
+
+				log.Infof("Retrying connection to %s in "+
+					"%s", c.cfg.Host, connectionRetryInterval)
+				time.Sleep(connectionRetryInterval)
+				continue reconnect
+			}
+
+			//Switch the new client with the old, shutdown one.
```
JoeGruffins: comment space

JoeGruffins:
@jholdstock this is working as expected for me. Let me know if you can find any holes, and if the intervals are ok. we could check for a disconnect less in my opinion.

jholdstock:
OK I've found something which seems a little funny, it might be nothing, but worth investigating.

Steps to reproduce:
- go install dcrd release-v.1.4.0
- go install dcrwallet master
- go install dcrstakepool/stakepoold master
- run harness

dcrd logs:
```
2019-08-30 13:27:59.840 [INF] RPCS: New websocket client 127.0.0.1:55288
2019-08-30 13:28:00.760 [INF] RPCS: New websocket client 127.0.0.1:55290
2019-08-30 13:28:04.894 [INF] RPCS: New websocket client 127.0.0.1:55292
2019-08-30 13:28:05.812 [INF] RPCS: New websocket client 127.0.0.1:55294
2019-08-30 13:28:09.949 [INF] RPCS: New websocket client 127.0.0.1:55296
2019-08-30 13:28:10.848 [INF] RPCS: New websocket client 127.0.0.1:55298
2019-08-30 13:28:15.009 [INF] RPCS: New websocket client 127.0.0.1:55300
2019-08-30 13:28:15.889 [INF] RPCS: New websocket client 127.0.0.1:55302
2019-08-30 13:28:20.047 [INF] RPCS: New websocket client 127.0.0.1:55304
2019-08-30 13:28:20.944 [INF] RPCS: New websocket client 127.0.0.1:55306
2019-08-30 13:28:25.122 [INF] RPCS: New websocket client 127.0.0.1:55308
2019-08-30 13:28:25.992 [INF] RPCS: New websocket client 127.0.0.1:55310
2019-08-30 13:28:30.153 [INF] RPCS: New websocket client 127.0.0.1:55312
2019-08-30 13:28:30.153 [INF] RPCS: Max websocket clients exceeded [25] - disconnecting client 127.0.0.1:55312
2019-08-30 13:28:31.048 [INF] RPCS: New websocket client 127.0.0.1:55314
2019-08-30 13:28:31.048 [INF] RPCS: Max websocket clients exceeded [25] - disconnecting client 127.0.0.1:55314
2019-08-30 13:28:35.206 [INF] RPCS: New websocket client 127.0.0.1:55316
2019-08-30 13:28:35.206 [INF] RPCS: Max websocket clients exceeded [25] - disconnecting client 127.0.0.1:55316
2019-08-30 13:28:36.097 [INF] RPCS: New websocket client 127.0.0.1:55318
2019-08-30 13:28:36.097 [INF] RPCS: Max websocket clients exceeded [25] - disconnecting client 127.0.0.1:55318
2019-08-30 13:28:40.249 [INF] RPCS: New websocket client 127.0.0.1:55320
2019-08-30 13:28:40.249 [INF] RPCS: Max websocket clients exceeded [25] - disconnecting client 127.0.0.1:55320
2019-08-30 13:28:41.158 [INF] RPCS: New websocket client 127.0.0.1:55322
2019-08-30 13:28:41.158 [INF] RPCS: Max websocket clients exceeded [25] - disconnecting client 127.0.0.1:55322
2019-08-30 13:28:45.307 [INF] RPCS: New websocket client 127.0.0.1:55326
2019-08-30 13:28:45.307 [INF] RPCS: Max websocket clients exceeded [25] - disconnecting client 127.0.0.1:55326
2019-08-30 13:28:46.201 [INF] RPCS: New websocket client 127.0.0.1:55328
2019-08-30 13:28:46.201 [INF] RPCS: Max websocket clients exceeded [25] - disconnecting client 127.0.0.1:55328
```

jholdstock:
This also seems to happen in dcrstakepool master and not related to this PR. Will open an issue to investigate further

JoeGruffins:
@jrick Thank you for your comments on the issue that led to this pr. Just wanted to know if you see any problems with this implementation.

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,127 @@
...
+
+			// Properly shutdown old client.
+			c.Shutdown()
+			c.WaitForShutdown()
+
+			// Start a new client.
+			client, err := rpcclient.New(c.cfg, c.ntfnHandlers)
+
+			if err != nil {
+				log.Infof("Failed to connect to %s: %v",
```
jholdstock: Should probably log at Warn level here.

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,127 @@
...
+					c.cfg.Host, err)
+
+				log.Infof("Retrying connection to %s in "+
+					"%s", c.cfg.Host, connectionRetryInterval)
+				time.Sleep(connectionRetryInterval)
+				continue reconnect
+			}
+
+			// Switch the new client with the old, shutdown one.
+			c.Client = client
```
jholdstock: Do we need a mutex on c.Client to make this safe for concurrent access?

JoeGruffins: probably? And then this should be changed to copy the rpcclient value?  I don't know what would happen if we took the address before a change and used it after. Panic? I was thinking to use a rwmutex here, use RLock for all commands and lock here, to let the client deal with the concurrency related to commands.

JoeGruffins: or does passing an address keep the old client alive until everyone is done with it? In which case maybe this is ok. I'll make some tests to see what happens. This is basic stuff I should know anyway.

JoeGruffins: I think it's ok as an address. https://play.golang.org/p/110tx7abKJa
Does the address need a lock?

chappjc: I've not reviewed this PR, but the linked play is a race.

chappjc: > or does passing an address keep the old client alive until everyone is done with it? In which case maybe this is ok. I'll make some tests to see what happens. This is basic stuff I should know anyway.

As long as the old client is referenced by a pointer somewhere then it a valid object.  But that doesn't mean concurrent read/write of `c.Client` is OK.  Here you write to it, but in another goroutine there is access to `c.Client.  You do not want the possibility of the write occurring at the same time as the read (or another write for that matter)!

JoeGruffins:
Testing with the -race flag, I believe that there are no longer any concurrency issues.

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,142 @@
...
+	mux    sync.RWMutex
+
+	cfg          *rpcclient.ConnConfig
+	ntfnHandlers *rpcclient.NotificationHandlers
+	wg           sync.WaitGroup
+	stop         chan struct{}
+	disconnected chan struct{}
+}
+
+// Rpccleint allows access to the underlying rpcclient by providing a copy of
```
jrick: typo

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -16,15 +16,29 @@ const (
...
 	cfg          *rpcclient.ConnConfig
 	ntfnHandlers *rpcclient.NotificationHandlers
 	wg           sync.WaitGroup
 	stop         chan struct{}
 	disconnected chan struct{}
 }
 
+// Rpccleint allows access to the underlying rpcclient by providing a copy of
+// its address.
+func (c *Client) Rpcclient() *rpcclient.Client {
```
chappjc: I think a more idiomatic formatting of this function name would be `RPCClient` or `RpcClient`, but I would prefer the former.

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -16,15 +16,29 @@ const (
...
+	c.mux.RLock()
+	client := c.client
+	c.mux.RUnlock()
+	return client
```
chappjc: Nothing wrong with the above 4 lines, but IMO a cleaner solution is

```go
c.mux.RLock()
defer c.mux.RUnlock()
return c.client
``

**Comment by JoeGruffins** [comment]
@jrick @chappjc Thank you.

@jrick To be clear, you think it would be better to request/make changes to rpcclient in the dcrd repo? I think it's more of a feature than a bug. We want the option to receive errors during reconnection rather than the current behavior, which is to wait until the commands can be sent and resend.

**Comment by chappjc** [comment]
I've always felt that rpcclient needs notifications for disconnect and reconnect.  But one part of what jrick said pertains to the fact that you have no idea what notifications you missed while you were disconnected.  Maybe that doesn't make a difference for dcrstakepool (maybe it does), but when you do reconnect or get reconnected automatically, you'd want to take stock of what has changed during the time you were disconnected.  This is a big open issue for dcrdata in fact.

**Comment by JoeGruffins** [comment]
Have implemented all change requests and left a note that this solution also is temporary. This issue was made to track further changes that need to be made to rpcclient #512

**Comment by jrick** [comment]
The notifications were certainly the main issue I encountered with rpcclient in wallet.  Applications really need to handle the disconnect and reconnect themselves in some sane way to ensure that state remains valid.  Many issues begin to arise when you try to do it generically underneath them.

It's for a similar reason that I opposed adding round-robin support to rpcclient: your application may assume certain data exists based on prior calls, but that is no longer the case when you start querying a cluster.  Load balancers can cause similar issues if your application isn't prepared for it, as I'm sure you are aware.  And again back to the notifications, it would have broken all of the stateful ones.  The network state and connections are best known and handled directly by the application.

This can be more work, but the benefit is large.  For dcrwallet, I ended up making the wallet package the state of the application and implemented syncers on top of it which call mutator methods to extend the chain, add transactions, etc.  If the client connection is lost, the syncer closes but the wallet state is left functional, ready to be synced by a new syncer later on.  When it came time to add SPV support, this was possible by writing an alternative syncer instead of turning the wallet package into a spaghetti code disaster (more so than it already is).  I can't say if this is the right approach for dcrstakepool, but it's an example of how an application can be better designed by handling the (very common) issue of network loss directly rather than abstracting it away.

**Comment by jholdstock** [comment]
I think this is looking good but I noticed something unusual in the logging.

1. Start the test harness
1. Wait for wallets to sync and rescan
1. Kill dcrwallet-1
1. Observe stakepoold-1 logging a disconnect and attempts to re-connect
1. Restart dcrwallet-1
1. stakepoold logs **a lot** of "client has been shutdown" warnings (thousands)

```
2019-09-04 10:47:04.864 [ERR] RPCC: Websocket receive error from 127.0.0.1:20111: websocket: close 1006 (abnormal closure): unexpected EOF
2019-09-04 10:47:12.772 [WRN] GRPC: Failed to connect to 127.0.0.1:20111: dial tcp 127.0.0.1:20111: connect: connection refused
2019-09-04 10:47:12.772 [INF] GRPC: Retrying connection to 127.0.0.1:20111 in 5s
2019-09-04 10:47:17.773 [WRN] GRPC: Failed to connect to 127.0.0.1:20111: dial tcp 127.0.0.1:20111: connect: connection refused
2019-09-04 10:47:17.773 [INF] GRPC: Retrying connection to 127.0.0.1:20111 in 5s
2019-09-04 10:47:22.773 [WRN] GRPC: Failed to connect to 127.0.0.1:20111: dial tcp 127.0.0.1:20111: connect: connection refused
2019-09-04 10:47:22.773 [INF] GRPC: Retrying connection to 127.0.0.1:20111 in 5s
2019-09-04 10:47:23.060 [INF] GRPC: ProcessWinningTickets: height 258810 block 000000043eeb93b8a004ff85676a309b804841b207be38bfa2401cb424a2ebd2 duration 5.159µs newvotes 0 duplicatevotes 0 errors 0
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for 4e9af3074ed4c386fd8e8a6fa58e3ece500ae482d1c56150ea343d7ce2f095a2
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for de3d7b3b62e543b4439faabe13a7f55a14419708ae9245bfac77e30d6e03703b
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for d58505319d547df33c13efdbc29430108213d09b705aa0eb79cdb34f47510c71
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for 2d8b556c955c3831d5aa3316ab675eb75f07fed2173539ad5daf71c74fae544f
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for 213fb4dbe12a105904e4a125c93840cc8bf76354f7b3cd4709eadcb7c012212a
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for d34445e94a2465290fcc22906436e58d46a0808487d665ece62af1b84c3cc274
2019-09-04 10:47:23.107 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for 629f6d373b7fb6703fd8fb16593c867c34cfea04ed2eaf2947f6d9cb0f13cdf0
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for a0db32b29f95db1a74442f9d0f25bc06f827ba3c4413b1f8a4850128228d1df1
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for 6b5f727374969f50e2bb625a02de6cc54a5e09f5f8e2429a81c0507ed0dd8073
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for 275a7796aeed97498dba6e2515913252f76cd2a4accaea7031398b9ef31bf6c1
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for e82babecf92c5151ee3cb1c1d6e5b4e57b26dd0a96ee1d0c2d174972e28e5815
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for bc774578d25f81cada64d7fe630ccd691a68c4ddce68b26282c32c465733ca35
2019-09-04 10:47:23.106 [WRN] GRPC: unexpected GetTransaction error: 'the client has been shutdown' for bca7d17901db399969b333ff4b23922c9ef1b12425932950b3a71ca7f5c011c3
[...]
```

**Comment by JoeGruffins** [comment]
@jholdstock Thank you, that was indeed a bad mistake.  The new notifications were were sending commands on the old client. Can you try again?

**Comment by JoeGruffins** [comment]
This works, but I need some time to figure out why it's necessary :/

**Comment by JoeGruffins** [comment]
As soon as the wallet is connected we get a OnSpentAndMissedTickets notification from dcrd with a lot of tickets in it.  We are unable to process those tickets because the wallet isn't re-connected yet, it happens in-between reconnect intervals. I wonder why we don't see this behavior on initial start up.

**Comment by JoeGruffins** [comment]
how's this? Wait until the wallet is connected, then continue to go through with the notification callback.

File: backend/stakepoold/rpc/rpcserver/client.go [hunk 7]
```diff
@@ -0,0 +1,164 @@
...
+			c.mux.Lock()
+			c.client.Shutdown()
+			c.client.WaitForShutdown()
+			// Switch the new client with the old, shutdown one.
+			c.client = client
+			c.mux.Unlock()
+
+			// Close the connected channel so that all waiting
+			// processes can continue.
+			c.connectedMux.Lock()
```
JoeGruffins: I'm not sure this lock is necessary.

JoeGruffins:
Sorry, I'm done. Will rebase if looks ok...
Again, this is temporary. Working on a long term solution.

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,168 @@
...
+// its address.
+func (c *Client) RPCClient() *rpcclient.Client {
+	c.mux.RLock()
+	defer c.mux.RUnlock()
+	return c.client
+}
+
+// New creates a new Client and starts the automatic reconnection handler.
+// Returns an error if unable to construct a new rpcclient.
+func New(cfg *rpcclient.ConnConfig, ntfnHandlers *rpcclient.NotificationHandlers) (*Client, error) {
```
chappjc: `rpcserver.New` strikes me as a confusing way to construct a new RPC client.

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+			// Client is disconnected. Try to reconnect.
+			// Start blocking on connected chan if not already.
+			c.connectedMux.Lock()
+			select {
+			case <-c.connected:
+				c.connected = make(chan struct{})
+			default:
+			}
+			c.connectedMux.Unlock()
```
chappjc: For the above dozen or so lines, much of `Connected` is duplicated, so why not do instead:

```go
select {
case <-c.Connected():   // not connected
default:                // still connected
	continue
}
```

Although, the above demonstrates that the semantics of `Connected()` are somewhat confusing, so maybe consider a `func IsConnected() bool`.

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,168 @@
...
+			// Start a new client.
+			client, err := rpcclient.New(c.cfg, c.ntfnHandlers)
+
+			if err != nil {
+				log.Warnf("Failed to connect to %s: %v",
+					c.cfg.Host, err)
+
+				log.Infof("Retrying connection to %s in "+
+					"%s", c.cfg.Host, connectionRetryInterval)
+				time.Sleep(connectionRetryInterval)
```
chappjc: An exponential backoff here would be better to avoid using up file descriptors trying to connect at a constant rate.

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,168 @@
...
+			// Properly shutdown old client.
+			c.mux.Lock()
+			c.client.Shutdown()
+			c.client.WaitForShutdown()
+			// Switch the new client with the old, shutdown one.
+			c.client = client
+			c.mux.Unlock()
+
+			// Close the connected channel so that all waiting
+			// processes can continue.
```
chappjc: 80 char wrap of all comments for consistency

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,168 @@
...
+			close(c.connected)
+
+			log.Infof("Reestablished connection to RPC server %s",
+				c.cfg.Host)
+
+			// Break out of the reconnect loop back to wait for
+			// disconnect again.
+			break reconnect
```
chappjc: note: just `break`

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,168 @@
...
+		c.wg.Add(1)
+		c.autoReconnect()
+	}(c)
+	return c, nil
+}
+
+// Stop stops automatic reconnections.
+func (c *Client) Stop() {
+	c.stop <- struct{}{}
+	c.wg.Wait()
```
chappjc: best to document what this is waiting for (the autoreconnect loop to return)

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,168 @@
...
+		ntfnHandlers: ntfnHandlers,
+		stop:         make(chan struct{}),
+		connected:    make(chan struct{}),
+	}
+	// A closed connected channel indcates successfully connected.
+	close(c.connected)
+	go func(c *Client) {
+		c.wg.Add(1)
+		c.autoReconnect()
+	}(c)
```
chappjc: instead of this closure as a goroutine, how about

```go
c.wg.Add(1)
go c.autoReconnect()
```

But even if you do have a func here, it's not necessary to pass `c` as an argument since it will be captured by the closure.

backend/stakepoold/rpc/rpcserver/client.go
```diff
@@ -0,0 +1,168 @@
...
+// disconnected and the channel is not yet blocking, creates a new channel that
+// will be closed on a successful reconnect.
+func (c *Client) Connected() <-chan struct{} {
+	c.connectedMux.Lock()
+	defer c.connectedMux.Unlock()
+	if c.RPCClient().Disconnected() {
+		// Start blocking on connected chan if not already.
+		select {
+		case <-c.connected:
+			c.connected = make(chan struct{})
```
chappjc: Is this expected to every be the case? (A disconnected `rpcclient.Client`, but `c.connected` still closed indicating that `Client` thinks it is connected?

backend/stakepoold/rpc/rpcserver/client.go
JoeGruffins: Ah, because it's not a server.

backend/stakepoold/rpc/rpcserver/client.go
JoeGruffins: Yes.
If the connected channel was not opened by Connected(), it will be closed here. Which will be most of the time.

backend/stakepoold/rpc/rpcserver/client.go
JoeGruffins: The interval is consistent with what is being done in dcrwallet:
https://github.com/decred/dcrwallet/blob/51900c2bd053bcca091d030611b6ada482719220/dcrwallet.go#L447-L454

I was going for simple as far as the timing is concerned.  Or perhaps I don't understand the problem.  I will google a bit more.

backend/stakepoold/rpc/rpcserver/client.go
JoeGruffins: Isn't it?

JoeGruffins:
@chappjc 
Thank you so much for the review.  I have a couple questions on your comments if you have time.

backend/stakepoold/rpc/rpcserver/client.go
JoeGruffins: Stopped at 2019-09-13 10:39:29.601 because I need to do something else
no signs of what happened in that issue. Readings from top, df, and lsof all seem normal on both stakepoold and dcrstakepool.

chappjc: It's confusing that "file descriptors" also refer to handles on other os resources like network connections, but you'd get a count for a certain process by looking in `/proc/{pid}/fd`.

I can only hope that what we saw years ago was the result of a bug causing unused connections to hang around eating up such handles.  You might consider lowering the retry time to 250ms and monitoring the processes fd proc folder to see if they accumulate and are not released as expected.

JoeGruffins: good idea, I'll do that before going to sleep and see what happens.

google told me that lsof -p PID would show me a list of "file descriptors"

chappjc: Ah, yes lsof does, sorry.

chappjc: Regardless of whether this file descriptor issue is still an issue, a reconnect interval that increases like 1, 2, 4, 8, 16, 32, 64, 64, 64, 64, ... is pretty simple and not uncommon reconnect behavior.  That said, I'm sure it's fine just to try every 5 seconds too.

JoeGruffins: One hour at 250ms and there was no change in the number of file descriptors.

dcrd's rpcclient does have an increasing timer. I promise to make sure the permanent answer to stakepoold's connection problems includes it.

Issue #291, opened 2019-02-03
Pubkey based authentication


xaur:
Pubkey = login, signed login challenge = password.

Address' pubkey can be used to not spawn an extra pubkey for the account.

xaur:
https://github.com/raedahgroup/vspd/pull/4 is exploring this. Client will prove ownership of the ticket reward address to log in.

PR #514, opened 2019-09-04
add per-ticket API authentication scheme

itswisdomagain:
Current API routes are mounted on `/api/v1/*` and `/api/v2/*`; and authentication is carried out by providing an API access token in the `Authorization` request header field in the format:
`Authorization: Bearer <api key>`.

This PR adds a new API authentication scheme (`TicketAuth`) as an alternative to the current `Bearer` auth scheme. The added `TicketAuth` scheme performs per-ticket authentication before executing requested actions on protected API endpoints such as `/getpurchaseinfo`. It's worth mentioning that the new auth scheme works as expected with all existing API endpoints and the changes introduced in this PR do not break backwards compatibility.

The new auth scheme uses the wallet built-in message signing capability and signs a msg which content is the current timestamp, allowing old msgs to be invalidated by the server so they are resistant to replay attacks.

The format for the `TicketAuth` scheme is:
`Authorization: TicketAuth SignedTimestamp=1564531200, Signature=frJIUN8DYpKDtOLCwo, TicketHash=591b17ed03afc916f274669939924e12ed5ac90d8cc602172fb53237b8a20522`

Where;
- TicketAuth: is the name of the authentication scheme
- SignedTimestamp: is the unix timestamp at the time of request, i.e. the message that is signed by the wallet
- Signature (base64 string): the result of signing the timestamp above using the private key for the ticket reward address belonging to the user's wallet (see below)
- TicketHash: the hash of the ticket for which the requested operation is to be performed

This authorization header format matches that defined in [RFC2617](http://www.ietf.org/rfc/rfc2617.txt). See this [SO answer](https://stackoverflow.com/a/11420667/6086027).

#### User's ticket reward address
A ticket purchased with voting rights shared with a vsp has 2 `sstxcommitment` outputs for receiving staking reward. One of these addresses belongs to the vsp wallet, while the other belongs to the ticket owner's wallet. This second address belonging to the ticket owner is what is used in preparing the `Signature` passed in the authentication header as described above.

#### Recovering redeem scripts for restored wallets
This ticket-level authentication makes recovering redeem scripts for restored wallets much easier. Instead of logging in to the vsp server to get the redeem script for a ticket or api key for an account, a client app can simply make an API request to `/getpurchaseinfo` using the `TicketAuth` scheme to retrieve the redeem script for a ticket - provided the ticket was purchased with the vsp.

Resolves #291. Part fix for #274 and https://github.com/decredcommunity/issues/issues/100.

Issue #504, opened 2019-08-23
Derive a new fee address for every ticket.


JoeGruffins:
Currently there is a limit on the number of fee address we can have.  This is set in three places:

https://github.com/decred/dcrstakepool/blob/3e03e20d438eef43274fd5f0e33251359621e445/controllers/main.go#L47
https://github.com/decred/dcrstakepool/blob/3e03e20d438eef43274fd5f0e33251359621e445/backend/stakepoold/server.go#L33
https://github.com/decred/dcrstakepool/blob/3e03e20d438eef43274fd5f0e33251359621e445/sample-dcrwallet.conf#L4

This limit is there because dcrwallet needs a list of addresses to search for the stakepool owner's fee address that the voter has agreed to pay a fee to. However, work is being done on dcrwallet to allow arbitrary xpubs to be imported into the wallet.

https://github.com/decred/dcrwallet/pull/1471

After getting that flushed out, and adding logic in dcrwallet to search addresses in that account for fees, we can remove the limit, and give out a new payment address for every ticket.  
The motivation being to:

   1.  Improve privacy. Currently an outside observer can tell that two tickets belong to the same person if they are paying to the same fee address, as it always the second output, but they will not be able to do that after these changes.

   2. Remove the artificial maxusers ceiling.

JoeGruffins:
I may be going about this all wrong.  Are we trying to drop dcrwallet completely in the end?

JoeGruffins:
Closing as repo is deprecated.

PR #515, opened 2019-09-04
add /purchaseticket API endpoint for accountless ticket purchase

### LINKED ISSUE #274: Make email optional
Only 20 stakepools know email addresses of people holding some 50% tickets. I don't know exact percentage, but it is big enough to be a concern.

Email is obsolete, legacy, ancient, anti-privacy comm system. It must not be a requirement for new systems built in 21 century to support the new decentralized finance.

Stakepools shall allow what I call "first party registration", where one is not forced into _any_ 3rd party to sign up with a service. Let users have a choice on their account recovery.

Please add 'privacy' label.
Similar to decred/politeia#554.
Related to #259.

itswisdomagain:
This introduces a new API endpoint to make ticket purchase possible without user signup/login.

This PR paired with #514 resolves #274 (also resolves https://github.com/decredcommunity/issues/issues/100) - emails become optional and vsp services such as purchasing tickets (implemented here) and setting voting preferences (implemented in #514) can be used without creating user accounts on the vsp.

Accountless ticket purchase is done by making a POST request to `api/{v1 or v2}/purchaseticket`, passing `UserPubKeyAddr` in post body. [This doc](https://docs.decred.org/wallets/cli/dcrwallet-tickets/#vsp-voting) describes how to generate a `pubkeyaddr` using `dcrwallet` and `dcrctl`.

Calling the new `/purchaseticket` endpoint with a valid `pubkeyaddr` returns the following (sample) data. Calling the endpoint with the same `pubkeyaddr` multiple times returns the same data. The message field in the response will read `APIPurchaseTicket: purchaseinfo retrieved for existing userPubKeyAddr`.

```
{
    "status": "success",
    "code": 0,
    "message": "APIPurchaseTicket: purchaseinfo generated for userPubKeyAddr",
    "data": {
        "PoolAddress": "TsoWCv4bDemoAddress",
        "PoolFees": 7.5,
        "Script": "512102ffbb6...sample script...efc8b552ae",
        "TicketAddress": "TcZJ5jMFDemoAddress",
        "VoteBits": 1,
        "VoteBitsVersion": 0
    }
}
```

The redeem script returned with the API response should be imported into the purchasing wallet so you can recover your funds and vote in the unlikely event of a VSP failure.

The ticket address, fee address and fee percentage should be used to create and broadcast the ticket purchase transaction. [Read how to do this with dcrwallet+dcrctl](https://docs.decred.org/wallets/cli/dcrwallet-tickets/#ticket-purchasing).

If you subsequently restore your wallet from seed, redeem scripts for "vsp tickets" would need to be (re)imported. Currently, a user can login to the vsp server to retrieve the redeem script. Since tickets purchase using this new API do not have user account login, redeem scripts can be retrieved by issuing an API call to `/getpurchaseinfo` using the ticket authentication scheme introduced in #514.


controllers/main.go
```diff
@@ -298,66 +300,61 @@ func (controller *MainController) APIAddress(c web.C, r *http.Request) ([]string
...
-		userFeeAddr.EncodeAddress(), importedHeight)
-
-	log.Infof("successfully create multisigaddress for user %d", c.Env["APIUserID"])
+	// load saved user info from db to get the user id
+	user = models.GetUserByEmail(dbMap, userPubKeyAddr)
 
-	err = controller.StakepooldUpdateUsers(dbMap)
-	if err != nil {
-		log.Warnf("failure to update users: %v", err)
+	if purchaseInfo := controller.generateTicketPurchaseInfo(dbMap, user, userPubKeyAddr); purchaseInfo != nil {
```
JoeGruffins: I would prefer this fail on error like most other stuff, but maybe just me.

itswisdomagain:
> Decrediton will have to use this api, maybe get their opinion?

Decrediton can use the API for sure, it's pretty similar to what Decrediton currently does, only that an API key isn't necessary.

As described here https://github.com/decred/decrediton/pull/280, to connect a vsp, a user provides the vsp host address and API Key. When a user submits these,
- Decrediton makes a GET request to `/api/v{n}/getpurchaseinfo` to confirm if the user has previously set up his vsp account by submitting a pubkey address.
- If the user hasn't done so, Decrediton generates a pubkey address from the user's wallet and submits the pubkey address in a POST request to `/api/v{n}/address`. The VSP uses the user's pubkey address to generate a multisig script, voting address and fee address for the owner of the API key.
- After submitting the pubkey address successfully, Decrediton makes another GET request to `/api/v{n}/getpurchaseinfo` to retrieve the msa and other data it needs for tickets purchase. This data is gotten from the first step if the user had already submitted a pubkey address.
- Decrediton saves the msa and other data gotten from `/api/v{n}/getpurchaseinfo` and uses it subsequently for tickets purchase.

It's worth noting that Decrediton has to authorize each of the above API using a valid API key; and to get this API key, the user first needs to create an account on the vsp frontend using an email address.

With this new API endpoint, Decrediton would no longer need an API key, just the vsp host address. Without requiring an API key, a user absolutely does not need to create an account from the vsp frontend and can purchase tickets directly from Decrediton.

Additionally, Decrediton does not need to make multiple API calls to get the information it needs to purchase tickets. This new API endpoint, both submits the user's pubkey address and returns the data needed by a client app such as Decrediton to complete the ticket purchase.

Also, with this new API endpoint, Decrediton can purchase multiple tickets without reusing voting or vsp fee addresses. Each purchase can begin with generating a pubkey address from the user's wallet, submitting that address to this new API endpoint which would return a new set of voting and vsp fee address that can be used for ticket purchase. Of course, Decrediton can, if it prefers, chose to save the first voting + vsp fee addresses and reuse those for subsequent ticket purchases.

> One concern I have is that it leaves cli users out of the loop. What about them? 

That's correct. Wasn't sure if it is necessary to provide support for cli users, if that is the case, such provision can be made. Cli support is currently possible because of the vsp frontend where users can submit their pubkey address and retrieve their multisig script+voting address+fee address.

> I can't say that this closes #274, or it's partner #291 without accounting for them.

Can't say this resolves #291 as #291 appears to refer to vsp login from the website. It does close #274 since with this PR, tickets can be purchased without using an email. #514 makes it possible to set voting preferences for tickets purchased this way also without requiring email. Email becomes optional - users can still chose to create accounts from the vsp frontend by providing an email address or use a client app like Decrediton to directly access vsp features without creating an account on the vsp frontend.

> And some things to consider going forward are that this area may change in a big way. We want to start producing a new redeem script for every ticket, and a new fee address.
> The fee addresses has an issue going #504
> The redeem script not yet. However the consensus is that we need the client to be able to reproduce the script. So, in the end the pool will need to import some more keys for every user, in order to make unique script hashes that are reproduce-able by the user.
> After that the pool fee will need to be paid out-of-band, in order to get coinjoin working according to jrick.
> 
> But for now this does look like a good system to me. Even after we get more privacy, a txid will still be a txid and the reward address will still be visible. Right? I still need to actually test this and the other, but if they work, I think this is progress in the right direction.

I agree with that last bit. This PR in no way prevents further work especially wrt adding the ability to (re)produce the redeem script on both the client and vsp end and improvements relating to payments of vsp fees.

What this PR does is simply make it possible for users to interact with/use vsp features without NEEDING to use an email address to create a user account on the vsp site.

controllers/main.go
```diff
@@ -743,95 +740,112 @@ func validateUserPubKeyAddr(pubKeyAddr string) (dcrutil.Address, error) {
...
-	userPubKeyAddr := r.FormValue("UserPubKeyAddr")
-
-	log.Infof("Address POST from %v, pubkeyaddr %v", remoteIP, userPubKeyAddr)
-
-	if _, err := validateUserPubKeyAddr(userPubKeyAddr); err != nil {
-		session.AddFlash(err.Error(), "address")
-		return controller.Address(c, r)
-	}
+func (controller *MainController) generateTicketPurchaseInfo(dbMap *gorp.DbMap, user *models.User,
+	userPubKeyAddr string) *poolapi.PurchaseInfo {
```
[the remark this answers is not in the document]
itswisdomagain: Similar to your previous comment, will implement. What I did was mostly a refactor of existing code and I noticed that some errors were not being logged, but returning an error will cause all errors to be logged - which is better I think.

JoeGruffins:
I've made a consumer of this api endpoint:
https://github.com/decred/tinydecred/pull/22
It seems to be working fine.

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/decred/dcrd/chaincfg/chainhash v1.0.2
 	github.com/decred/dcrd/chaincfg/v2 v2.3.0
 	github.com/decred/dcrd/dcrec v1.0.0
+	github.com/decred/dcrd/dcrec/edwards v1.0.0 // indirect
```
JoeGruffins: It doesn't look like you've added any imports, perhaps you can remove this?

controllers/main.go
```diff
@@ -301,66 +303,64 @@ func (controller *MainController) APIAddress(c web.C, r *http.Request) ([]string
...
-	var importedHeight int64
-	importedHeight, err = controller.Cfg.StakepooldServers.ImportNewScript(serializedScript)
-	if err != nil {
-		return nil, codes.Unavailable, "system error", errors.New("unable to process wallet commands")
+	user = &models.User{
+		Username:        userPubKeyAddr,
+		Email:           userPubKeyAddr,
+		EmailVerified:   0,
+		VoteBits:        1,
+		VoteBitsVersion: int64(controller.voteVersion),
```
JoeGruffins: Perhaps it would be best to put in some random password? Just to be safe and sure noone can access this account through http...

xaur:
(1) Will `purchaseticket` result code/status be different for newly registered VSP ticket purchase and just reporting existing one?

(2) Is `purchaceticket` protected by authentication in #514? (I assume yes, otherwise it is possible to learn and manipulate others' tickets)

(3) How voting prefs are set and changed? Can you provide or point to examples of all requests and responses?

(4) In the current system the communication with the VSP is done once via the browser, and later the user just broadcasts VSP ticket transactions and the VSP detects them. In the new system, each new pubkeyaddr requires communicating with the VSP. Forcing users to use the website is bad anyway (I'd like to do everything without browsers), but if new address is used for each tickets, using the website simply becomes infeasible. This means changes to dcrwallet or a separate staking program is required for CLI users. I would prefer the latter to avoid introducing VSP protocols into dcrwallet.

(5) In case new address is used for each ticket purchase, redeem script returned for _each_ ticket needs to be backed up (!) by the client and imported into his wallet. This will require automation and new features on the client side - either changes to dcrwallet (including commands to export and import an ever growing bundle of redeem scripts) or a standalone program. Server side also needs tools to backup/restore all the redeem scripts. Without these, the new APIs will be unusable (in no address reuse mode). Is that correct?

> if it is necessary to provide support for cli users

Absolutely necessary. The approach taken everywhere else in Decred is that CLI support is added _first_, and then some of that is reused in GUIs.

Also note that "real" CLI support means all important operations can be done without ever using the ugly web browsers.

> #291 appears to refer to vsp login from the website

Per the above, ideally it refers to all kinds of access including non-website.

xaur:
The whole problem with redeem scripts outlined in (5) goes away if https://github.com/decred/dcrwallet/issues/1613 is implemented.

JoeGruffins:
Long time no see! Thank you for continuing this!

I have started adding tests to main.go. It would be great if you could add tests for generateTicketPurchaseInfo. It should be done on top of #602 There you should have all the tools you need to set up a mock stakepoold and sql.

I also still think you need to throw a random password in there for the new accountless users.

JoeGruffins:
One concern.

Right now new users have the hurdle of a captcha that supposedly prevents automation of creating new users. With this change, what is to prevent a hostile user from creating new users until we hit max users? Suddenly hitting max users would be pretty catastrophic for the stakepool owner, and the attacker could continue if they up the max. Eventually the database would take up lots of space, and poolfee and voting addresses could be exhausted.

Either we need a puzzle here, or not save these users until they buy a ticket, which creates difficulties, because then how do we watch for tickets unless we watch even bogus users?

Or we need to be informed of ticket purchases by these users, then confirm that they have indeed bought said ticket before we add and watch them.

Even if they are all good users, this may bring us to max users much faster if users make a new account for every ticket.

xaur:
How does this work compare to https://github.com/decred/dcrstakepool/issues/574 ? Is there any cooperation? (I'm concerned about wasted efforts)

dajohi:
This is a PR and #574 is an issue

itswisdomagain:
Closing in favour of #625

PR #520, opened 2019-09-10
enablevoting=0 in dcrwallet conf

jholdstock:


JoeGruffins:
Is the redundancy a problem?

jholdstock:
The problem is that enablevoting=1 will cause dcrwallet to vote without respecting the users votebits. stakepoold does the voting with the correct votebits.

JoeGruffins:
That's a pretty big deal.

itswisdomagain:
@jholdstock Is there a way to ensure at runtime that this dcrwallet config value is correctly set? It could be so easy to miss resulting in the issue you described above.

JoeGruffins:
@itswisdomagain can there be different votes for different tickets that way?

itswisdomagain:
Missed that @JoeGruffins. It cannot, currently.
> @itswisdomagain can there be different votes for different tickets that way?

So I guess we look for a way to ensure at runtime that the dcrwallet `enablevoting` config value is correctly set.

JoeGruffins:
We can get it through walletinfo https://github.com/decred/dcrwallet/blob/51900c2bd053bcca091d030611b6ada482719220/rpc/jsonrpc/methods.go#L3477

Issue #518, opened 2019-09-09
Add errors package


JoeGruffins:
This is just a suggestion to get our errors more organized.  There are many times when we log errors twice, probably some places where errors aren't logged at all.  We could start organizing them better with a scheme descriped here: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully
This is also something that is being done in dcrwallet: https://github.com/decred/dcrwallet/tree/master/errors

JoeGruffins:
A valuable opinion on the topic 
https://matrix.to/#/!HEeJkbPRpAqgAwhXWO:decred.org/$156882789034142KmUIm:decred.org
tldr: It's too complex.

jholdstock:
Because we support both the current and the previous versions of Go, dcrwallet has had to include a shim to make use of the new error features in 1.13 and still maintain compatibility with 1.12

https://github.com/decred/dcrwallet/blob/4bd4faa11f9b7ad0cb82a2107d22e7f1d023803b/errors/errors.go#L5-L7

We won't need to include this shim if we wait until Go 1.14 is released, which should make implementing your suggestion slightly less painful.

PR #524, opened 2019-09-12
Ensure low fee tickets are detected upon maturation.

### LINKED ISSUE #521: Low fee tickets isnt working
As reported by [@girino:decred.org on Matrix](https://matrix.to/#/!XijfkJigmGCmMkgTTl:decred.org/$156797243524684yfzhv:decred.org?via=decred.org&via=matrix.org), the low fee tickets feature is not working correctly.

Low fee tickets are showing in the users /tickets page, but not in the admin tickets page.

jholdstock:
The low fee tickets feature has been broken since inception.
There are multiple places where tickets are validated and
they had inconsistent error handling.

This PR also includes some logging and commenting I added
whilst investigating the issue.

Closes #521

### DIFF
--- a/backend/stakepoold/rpc/rpcserver/context.go
+++ b/backend/stakepoold/rpc/rpcserver/context.go
@@ -484,8 +484,13 @@ func (ctx *AppContext) vote(wg *sync.WaitGroup, blockHash *chainhash.Hash, block
 	w.sendDuration = time.Since(startSend)
 }
 
+// processNewTickets is invoked every time a new block is created. Any tickets
+// which matured in this block will be included in the parameter.
 func (ctx *AppContext) processNewTickets(nt NewTicketsForBlock) {
 	start := time.Now()
+
+	log.Debugf("processNewTickets: Block %d contains %d newly matured tickets",
+		nt.BlockHeight, len(nt.NewTickets))
 
 	// We use pointer because it is the fastest accessor.
 	newtickets := make([]*ticketMetadata, 0, len(nt.NewTickets))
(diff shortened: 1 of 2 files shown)


backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -552,30 +558,33 @@ func (ctx *AppContext) processNewTickets(nt NewTicketsForBlock) {
...
 	liveTicketsCount := len(ctx.LiveTicketsMSA)
 	ctx.Unlock()
-	log.Debug("processNewTickets ctx.Unlock")
+	log.Debug("processNewTickets: ctx.Unlock")
 
 	// Log ticket information outside of the handler.
 	go func() {
 		for ticket, msa := range newLiveTickets {
-			log.Infof("added new live ticket %v msa %v", ticket, msa)
+			log.Infof("processNewTickets: added new live ticket %v msa %v", ticket, msa)
```
JoeGruffins: If this is info, the average person might not know what an msa is. Could we write it out? I see you do write it out later.

backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -521,21 +526,22 @@ func (ctx *AppContext) processNewTickets(nt NewTicketsForBlock) {
 
 		msgTx, err := MsgTxFromHex(n.hex)
 		if err != nil {
-			log.Warnf("MsgTxFromHex failed for %v: %v", n.hex, err)
+			log.Warnf("processNewTickets: MsgTxFromHex failed for %v: %v", n.hex, err)
```
JoeGruffins: If we end up adding an error package like suggested in #518 , we might also start using the convention in dcrwallet where the operation is declared at the beginning of a function:
const op = "processNewTickets"
Then when we make an error
err := fmt.Errorf("%v: everybody died", op)
Just a suggestion tho, we can always decide later when/if that issue gets worked on.

backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -552,30 +558,33 @@ func (ctx *AppContext) processNewTickets(nt NewTicketsForBlock) {
...
 	// Log ticket information outside of the handler.
 	go func() {
 		for ticket, msa := range newLiveTickets {
-			log.Infof("added new live ticket %v msa %v", ticket, msa)
+			log.Infof("processNewTickets: added new live ticket %v msa %v", ticket, msa)
 		}
 
 		for ticket, msa := range newIgnoredLowFeeTickets {
-			log.Infof("added new ignored ticket %v msa %v", ticket, msa)
+			log.Infof("processNewTickets: added new ignored ticket %v msa %v", ticket, msa)
```
JoeGruffins: msa here too

backend/stakepoold/rpc/rpcserver/context.go
```diff
@@ -521,21 +526,22 @@ func (ctx *AppContext) processNewTickets(nt NewTicketsForBlock) {
...
+			newIgnoredLowFeeTickets[*n.ticket] = n.msa
 			log.Warnf("ignoring ticket %v for msa %v ticketFeesValid %v err %v",
 				n.ticket, n.msa, ticketFeesValid, err)
-			newIgnoredLowFeeTickets[*n.ticket] = n.msa
 		}
-
-		newLiveTickets[*n.ticket] = n.msa
 	}
 
-	log.Debug("processNewTickets ctx.Lock")
```
chappjc: I think these lock/unlock debug logs can be removed

JoeGruffins:
It works!

backend/stakepoold/rpc/rpcserver/context.go
jholdstock: We can go through and do this later in #518. I might actually make a start on that this afternoon

jholdstock:
Thanks all for the reviews. I have implemented all suggestions apart from the error change from Joe, I will implement that in #518 

> after adding the tickets, they still show as invalid for the user. Do you see the same?

Yeah I see this as well. This was how the original implementation worked. We could possibly change this later, in another PR, but I don't think it is worth investing much more time into this feature. Having spoken with VSP admins, it seems like the feature is not really used since VSP staking was added to decrediton.

PR #526, opened 2019-09-13
Compare ColdWalletExtPub config value across dcrstakepool and all stakepoold configs

### LINKED ISSUE #450: Check that all coldwalletextpub are the same accross all config files

itswisdomagain:
Check that coldwalletextpub in `dcrstakepool.conf` matches that in all `stakepoold.conf`s.
Resolves #450.


server.go
```diff
@@ -130,6 +130,21 @@ func runMain() error {
...
+
+	// Check that coldwalletextpub is the same in dcrstakepool.conf and all stakepoold.conf files
+	for _, stakepooldColdWalletResponse := range stakepooldColdWalletResponses {
+		if cfg.ColdWalletExtPub != stakepooldColdWalletResponse.ColdWalletExtPub {
+			err := fmt.Errorf("coldwalletextpub is not the same in dcrstakepool.conf and stakepoold.conf files")
+			fmt.Fprintln(os.Stderr, err)
+			return err
+		}
+	}
+
```
JoeGruffins: Can you combine both of these into one method on stakepooldclient that takes the xpub and gives back ok or not?

stakepooldclient/stakepooldclient.go
```diff
@@ -704,3 +704,17 @@ func (s *StakepooldManager) GetStakeInfo() (*pb.GetStakeInfoResponse, error) {
...
 	return nil, errors.New("GetStakeInfo RPC failed on all stakepoold instances")
 }
+
+func (s *StakepooldManager) GetAllColdWalletExtPub() ([]*pb.GetColdWalletExtPubResponse, error) {
+	responses := make([]*pb.GetColdWalletExtPubResponse, len(s.grpcConnections))
+	for i, conn := range s.grpcConnections {
+		client := pb.NewStakepooldServiceClient(conn)
+		resp, err := client.GetColdWalletExtPub(context.Background(), &pb.GetColdWalletExtPubRequest{})
+		if err != nil {
+			log.Warnf("GetColdWalletExtPub RPC failed on stakepoold instance %s: %v", conn.Target(), err)
```
JoeGruffins: I think this should be Errorf, as we cannot continue.  But really you can pass it on as this will log twice (three times actually).

server.go
dajohi: agree

server.go
```diff
@@ -126,8 +126,16 @@ func runMain() error {
...
 		application.Close()
-		return fmt.Errorf("Failed to initialize the main controller: %v",
-			err)
+		return fmt.Errorf("Failed to initialize the main controller: %v", err)
+	}
+
+	// Check that dcrstakepool config and all stakepoold configs
+	// have the same value set for `coldwalletextpub`.
+	if !controller.Cfg.StakepooldServers.ColdWalletExtPubMatches(cfg.ColdWalletExtPub) {
+		err := fmt.Errorf("coldwalletextpub is not the same in " +
```
JoeGruffins: You don't have to log here, it will be logged after returned.

Currently:
Untitled

Also, we have only been wrapping comments to 80 columns length, because it may go into godoc, I believe is the reason.  You don't have to wrap here, but I really don't mind if it's easier for you to read.

stakepooldclient/stakepooldclient.go
```diff
@@ -704,3 +704,18 @@ func (s *StakepooldManager) GetStakeInfo() (*pb.GetStakeInfoResponse, error) {
 	}
 	return nil, errors.New("GetStakeInfo RPC failed on all stakepoold instances")
 }
+
+func (s *StakepooldManager) ColdWalletExtPubMatches(dcrstakepoolColdWalletExtPub string) bool {
```
JoeGruffins: Can you add a comment above this method starting with the method name?

server.go
itswisdomagain: Yeah, I find shorter-length comments easier to read.

JoeGruffins:
Needs to be bumped in two places, second is here:
https://github.com/decred/dcrstakepool/blob/185f8f201856f932c0d39653e668feeaf436f9ff/stakepooldclient/stakepooldclient.go#L23

stakepooldclient/stakepooldclient.go
```diff
@@ -704,3 +704,23 @@ func (s *StakepooldManager) GetStakeInfo() (*pb.GetStakeInfoResponse, error) {
 	}
 	return nil, errors.New("GetStakeInfo RPC failed on all stakepoold instances")
 }
+
+// ColdWalletExtPubMatches calls GetColdWalletExtPub RPC on all stakepoold
+// instances and compares the returned `coldwalletextpub` value against the
+// value set in dcrstakepool's config.
+// Returns false if an RPC call to any of the backend clients errors or
+// if any returned `coldwalletextpub` value is not the same as dcrstakepool's.
+func (s *StakepooldManager) ColdWalletExtPubMatches(dcrstakepoolColdWalletExtPub string) bool {
```
jholdstock: I think this func should return `(bool, error)` and the calling code should have distinct logging for cases of RPC failure, and cases of incorrect coldwalletextpub config.

Right now if the RPC call fails (eg. no network) then we will log both messages:

`GetColdWalletExtPub RPC failed on stakepoold instance %s: %v`
  followed by
`coldwalletextpub is not the same in dcrstakepool.conf and stakepoold.conf files"`


We should also include the `ip:port` of the incorrectly configured stakepoold in the log message

itswisdomagain: Is the latest implementation better?

jholdstock: Looks good, thanks.

PR #527, opened 2019-09-13
multi: cleanup

dajohi:


system/core.go
```diff
@@ -38,8 +38,7 @@ func GojiWebHandlerFunc(h http.HandlerFunc) web.HandlerFunc {
 
 func (application *Application) Init(APISecret string, baseURL string,
```
itswisdomagain: `APISecret, baseURL, cookieSecret string, cookieSecure bool, DBHost, DBName, DBPassword, DBPort, DBUser string`?

controllers/main.go
```diff
@@ -226,6 +225,7 @@ func dcrDataAgendas(url string) ([]*dcrdatatypes.AgendasInfo, error) {
 		return nil, err
 	}
 	data, err := ioutil.ReadAll(resp.Body)
+	resp.Body.Close()
```
JoeGruffins: oops

controllers/main.go
```diff
@@ -1258,10 +1270,9 @@ func (controller *MainController) PasswordResetPost(c web.C, r *http.Request) (s
 	if !controller.IsCaptchaDone(c) {
 		session.AddFlash("You must complete the captcha.", "passwordresetError")
 		return controller.PasswordReset(c, r)
-	} else {
-		session.Values["CaptchaDone"] = false
-		c.Env["CaptchaDone"] = false
 	}
+	session.Values["CaptchaDone"] = false
```
JoeGruffins: @amassari also noticed this in #491

controllers/main.go
```diff
@@ -1210,16 +1214,24 @@ func (controller *MainController) Index(c web.C, r *http.Request) (string, int)
 	t := controller.GetTemplate(c)
 
 	// execute the named template with data in c.Env
-	widgets := helpers.Parse(t, "home", c.Env)
-
+	widgets, err := helpers.Parse(t, "home", c.Env)
+	if err != nil {
+		log.Infof("helpers.Parse: home failed: %v", err)
+		return "/error?r=/stats", http.StatusSeeOther
+	}
```
JoeGruffins: This changes behavior, but it looks reasonable.

server.go
```diff
@@ -276,7 +276,11 @@ func runMain() error {
 	graceful.PostHook(func() {
 		application.Close()
 	})
-	app.Abandon(middleware.Logger)
+	err = app.Abandon(middleware.Logger)
+	if err != nil {
+		return fmt.Errorf("could not Abandon logger: %v", err)
+	}
+
```
JoeGruffins: This also changes behavior

controllers/main.go
```diff
@@ -1210,16 +1214,24 @@ func (controller *MainController) Index(c web.C, r *http.Request) (string, int)
 	t := controller.GetTemplate(c)
 
 	// execute the named template with data in c.Env
-	widgets := helpers.Parse(t, "home", c.Env)
-
+	widgets, err := helpers.Parse(t, "home", c.Env)
+	if err != nil {
+		log.Infof("helpers.Parse: home failed: %v", err)
```
jholdstock: Log error here instead of info

controllers/main.go
```diff
@@ -1210,16 +1214,24 @@ func (controller *MainController) Index(c web.C, r *http.Request) (string, int)
...
+		return "/error?r=/stats", http.StatusSeeOther
```
jholdstock: We removed the functionality to redirect from the error page. Param `?r=/stats` is not used anywhere and can be removed.

I notice this is present in a few other places, not related to your PR. Would you mind removing them please?

controllers/main.go
```diff
@@ -1210,16 +1214,24 @@ func (controller *MainController) Index(c web.C, r *http.Request) (string, int)
...
 	c.Env["IsIndex"] = true
 	c.Env["Title"] = "Decred Voting Service - Welcome"
 	c.Env["Designation"] = controller.Cfg.Designation
 
 	c.Env["Content"] = template.HTML(widgets)
 
-	return helpers.Parse(t, "main", c.Env), http.StatusOK
+	doc, err := helpers.Parse(t, "main", c.Env)
+	if err != nil {
+		log.Infof("helpers.Parse: main failed: %v", err)
```
jholdstock: Same 2 comments as above - remove `?r=` param and log level error

controllers/main.go
jholdstock: Agreed, tested this out and the new behaviour is an improvement.

PR #528, opened 2019-09-14
stakepoold: Add shutdown context

JoeGruffins:
A struct channel was used in order to signal shutdowns. This changes
that to a context.Context that we can spawn new contexts from. Abstracts
the task to a seperate file signal.go

This is pretty much a copy/paste from dcrd and dcrwallet.  This gives us a base context to spawn new contexts off of.

One difference is that the name "ctx" is taken. Can we rename all "ctx" to something else?

Also, no idea what to do with the copyrights. Help me out.

### DIFF
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -5,15 +5,16 @@
 package main
 
 import (
+	"context"
 	"encoding/gob"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
-	"os/signal"
 	"path/filepath"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/decred/dcrd/chaincfg"
@@ -117,7 +118,10 @@ func deriveChildAddresses(key *hdkeychain.ExtendedKey, startIndex, count uint32,
 	return addresses, nil
 }
 
-func runMain() error {
+func runMain(shutdownContext context.Context) error {
+	// WaitGroup to pass around and wait, after shutdown signal is received,
+	// for goroutines to safely stop.
+	wg := new(sync.WaitGroup)
 	// Load configuration and parse command line.  This function also
 	// initializes logging and configures it accordingly.
 	loadedCfg, _, err := loadConfig()
@@ -155,7 +159,7 @@ func runMain() error {
 	rpcclient.UseLogger(clientLog)
 
 	var walletVer semver
-	walletConn, walletVer, err := connectWalletRPC(cfg)
+	walletConn, walletVer, err := connectWalletRPC(shutdownContext, wg, cfg)
 	if err != nil || walletConn == nil {
 		log.Infof("Connection to dcrwallet failed: %v", err)
 		return err
@@ -214,7 +218,6 @@ func runMain() error {
 		PoolFees:               cfg.PoolFees,
 		NewTicketsChan:         make(chan rpcserver.NewTicketsForBlock),
 		Params:                 activeNetParams.Params,
-		Quit:                   make(chan struct{}),
 		SpentmissedTicketsChan: make(chan rpcserver.SpentMissedTicketsForBlock),
 		UserData:               userData,
 		UserVotingConfig:       userVotingConfig,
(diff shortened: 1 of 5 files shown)


backend/stakepoold/signal.go
```diff
@@ -0,0 +1,54 @@
...
+package main
+
+import (
+	"context"
+	"os"
+	"os/signal"
+)
+
+// shutdownSignaled is closed whenever shutdown is invoked through an interrupt
+// signal or from an JSON-RPC stop request.  Any contexts created using
```
JoeGruffins: this json-rpc bit doesn't belong

backend/stakepoold/signalsigterm.go
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+// +build darwin dragonfly freebsd linux netbsd openbsd solaris
```
jholdstock: I asked you this question on Matrix, but reposting here for posterity.

> how did you identify that we need this? because I notice that has been newly added and there was nothing doing that before your PR

JoeGruffins: Was a copy paste from wallet/dcrd and had not looked into its necessity. Since this is meant to run on linux, and we already had the working and tested signal.Notify(c, os.Interrupt) I will just use that. If this init turns out to be necessary, it can easily be added later.

backend/stakepoold/server.go
```diff
@@ -338,27 +337,10 @@ func runMain() error {
...
-		close(ctx.Quit)
-	}()
-
 	ctx.Wg.Add(3)
-	go ctx.NewTicketHandler()
-	go ctx.SpentmissedTicketHandler()
-	go ctx.WinningTicketHandler()
+	go ctx.NewTicketHandler(shutdownContext.Done())
+	go ctx.SpentmissedTicketHandler(shutdownContext.Done())
+	go ctx.WinningTicketHandler(shutdownContext.Done())
```
itswisdomagain: Should be possible for each of these methods to create their own shutdownContext using `withShutdownCancel`, otherwise this isn't much different from what we currently have.

backend/stakepoold/server.go
```diff
@@ -377,18 +359,25 @@ func runMain() error {
 		}()
 	}
 
-	// Wait for CTRL+C to signal goroutines to terminate via quit channel.
+	// Wait for CTRL+C to signal goroutines to terminate
+	<-shutdownContext.Done()
```
itswisdomagain: If my previous comment is applied, we'd only need to `ctx.Wg.Wait()` here without waiting for the `shutdownContext` to be canceled.

Actually, waiting for this `shutdownContext` to be canceled isn't relevant either ways, since we'll still have to wait for the waitgroup counter to return to 0; and each of the goroutines that decrement the waitgroup counter do so only after the `shutdownContext` is canceled.

If there's something we should do here, is make the waitgroup counter increment whenever a shutdown context is created, then ensure that each method that creates and uses a shutdown context decrements the waitgroup after performing any necessary cleanups on receiving the context cancelation signal. Then here, we  only wait for the waitgroup counter to decrement to 0.

backend/stakepoold/server.go
JoeGruffins: It is the same. This is about readability and standardization.

itswisdomagain: Oh, I see. We could go the extra mile and have those 3 methods and any other method that requires a context, create their own shutdownContext using `withShutdownCancel`. What do you think?

JoeGruffins: hmmm, idk. Would that be better? I'll look a little more into when the other decred repo's might spawn a new context. When I say standardization, I mostly mean what wallet and dcrd are doing. I think there a lot of good ideas that have developed that we could make use of.  Although this particular refactor has been around since btcd apparently, just without the context. It was just a channel at first. 
https://github.com/decred/dcrwallet/commit/b145868a4bc5e8bd9fec82afc0a806e0d2529f67#diff-dca9e25a1ecb5a565b1bf9142c755f71

backend/stakepoold/signal.go
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2013-2014 The btcsuite developers
```
jholdstock: Add on line 2:

```
// Copyright (c) 2019 The Decred developers
```

JoeGruffins:
So, I went a little farther. Let me know what you think, I can always pop off the last commit.
Move waitgroup to the server.  This is done because It's the server waiting, and I think it works well to put the wg there.
Pass the waitgroup around by reference.  This isn't something I've seen done, but it makes sense to me.  The wg.Add(1) should be called right before whatever go func it's waiting on, and then that funk needs to notify when done.
If we make it a rule then we can take out that one line that was kinda useless as @itswisdomagain pointed out.

And then pass the context instead of the channel.  I'm kinda torn here, as we could just pass the channel, but it doesn't hurt to pass the entire context, does it? Anyway, I like it like this.  Also wires the autoreconnect to watch the shutdown context's done channel, and removes the need to call stop.

JoeGruffins:
Not sure what this travis error is all about.

chappjc:
Restarting the build won't help in this case. Go 1.13 now validates all pseudo version strings. gitea specified too much one dep's commit hash.

You'll have to put a replace statement:

```
replace github.com/go-macaron/cors v0.0.0-20190309005821-6fd6a9bfe14e9 => github.com/go-macaron/cors 6fd6a9bfe14e
```

JoeGruffins:
@chappjc Thank you!!

backend/stakepoold/server.go
jholdstock: With the latest changes which have been pushed, I think the wait for shutdownContext is now redundant and can be removed. There is no harm in keeping it there, but it is not necessary. Do you concur @itswisdomagain @JoeGruffins ?

PR #529, opened 2019-09-15
stakepoold: context -> stakepool

JoeGruffins:
I do not think that context.go belongs in the rpc package.  If it were just relaying rpc commands, sure, but it does and will do more.  The name "context" also conflicts with context.Context, which we want to start using to pass shutdowns and "request-scoped values across API boundaries and between processes"

This pr moves context to its own package named stakepool, renames the main struct Stakepool, and renames all references to it stakepool.

this work is on top of #528

### DIFF
--- a/backend/stakepoold/server_test.go
+++ b/backend/stakepoold/server_test.go
@@ -86,12 +86,12 @@ func randString(n int) string {
 }
 
 var (
-	c  *stakepool.Stakepoold
-	wt stakepool.WinningTicketsForBlock
+	spd *stakepool.Stakepoold
+	wt  stakepool.WinningTicketsForBlock
 )
 
 func init() {
-	c = &stakepool.Stakepoold{
+	spd = &stakepool.Stakepoold{
 		LiveTicketsMSA: make(map[chainhash.Hash]string),
 		VotingConfig: &stakepool.VotingConfig{
 			VoteBits:         1,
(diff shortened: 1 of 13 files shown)


JoeGruffins:
actually
AppContext -> Stakepoold
ctx -> spd

JoeGruffins:
Thanks for thumbs up guys! I wasn't sure how this would be accepted.  I'll go ahead and rebase if #528 goes through then.

backend/stakepoold/stakepool/stakepool.go
```diff
@@ -1,4 +1,8 @@
-package rpcserver
+// Copyright (c) 2019 The Decred developers
+
+//Package stakepool holds the Stakepool struct and processes incomming commands
```
JoeGruffins: spacing

backend/stakepoold/signal.go
```diff
@@ -19,13 +19,13 @@ var shutdownSignaled = make(chan struct{})
 // withShutdownCancel creates a copy of a context that is cancelled whenever
 // shutdown is invoked through an interrupt signal or from an JSON-RPC stop
 // request.
-func withShutdownCancel(ctx context.Context) context.Context {
-	ctx, cancel := context.WithCancel(ctx)
+func withShutdownCancel(spd context.Context) context.Context {
```
jholdstock: I think this should have remained `ctx context.Context`?

backend/stakepoold/server_test.go
```diff
@@ -86,14 +86,14 @@ func randString(n int) string {
 }
 
 var (
-	c  *rpcserver.AppContext
-	wt rpcserver.WinningTicketsForBlock
+	c  *stakepool.Stakepoold
```
jholdstock: Rename to `spd` for consistency?

backend/stakepoold/signal.go
JoeGruffins: Good catch.

backend/stakepoold/server_test.go
JoeGruffins: Even better catch.

go.mod
```diff
@@ -32,7 +32,6 @@ require (
 	github.com/gorilla/websocket v1.4.1 // indirect
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/jrick/logrotate v1.0.0
-	github.com/kr/pretty v0.1.0 // indirect
```
JoeGruffins: see, I switched to go 1.13 and It wants this change.

Issue #525, opened 2019-09-12
Prometheus metrics


isuldor:
Looking for feedback as I've added instrumentation to stakepoold in my environment to optionally export metrics for Prometheus. You can see the changes here:
https://github.com/isuldor/dcrstakepool/pull/1/files

Obviously it doesn't make any sense to merge this upstream if no one else is using Prometheus, as it has the downside of requiring three Prometheus client libraries. But the upside to using Prometheus to monitor stakepoold directly is that it can detect outages pretty fast (ie. giving you a chance to fix things before dcrstakepool implodes on itself). If there's interest, I can write a tutorial on setting up Prometheus to scrape stakepoold and send out alerts when necessary. I've been sending alerts to my mobile using pushover.net (5 bucks for life) and there is also opsgenie (like pagerduty, but free for up to 5 users). I might also look into deploying this stuff with Docker (ala lndmon), but that's further down the line for me.

So, would you want this PR for exporting Prometheus metrics?

Update: Oh, these libraries being imported also include methods for parsing or bridging into other time-series/monitoring systems like InfluxDB, Sensu, Metricbeat, Datadog, etc.  So an operator using any of those systems should be able to make use of these metrics. Apparently [Openmetrics](https://opensource.googleblog.com/2018/08/openmetrics-project-accepted-into-cncf.html) is going to be based on the Prometheus exposition format.  If no one has strong objections, I'll submit a PR next week.

PR #530, opened 2019-09-18
Optionally export metrics for external monitoring

isuldor:
WIP
- [x] use config.go for default values
- [x] change function to appcontext method
- [ ] shutdown channel
- [ ] wait group
- [ ] prometheus setup guide
- [ ] readme one-liner

I've added optional Prometheus instrumentation (`--prometheus`) for monitoring and outage alerts (#525).  The intention is to add reliability to a vsp by enabling the operator to quickly detect outages. An external monitoring system can be set up to monitor stakepoold instances, in addition to the existing api health check endpoint provided by dcrstakepool.  The advantage to detecting stakepoold outages is that they generally happen before dcrstakepool falls over.  This will likely become increasingly critical once the planned changes for [fee addresses](https://github.com/decred/dcrstakepool/issues/504) is implemented. The Prometheus client libraries also include parsing and bridges for other monitoring systems, such as InfluxDB, Sensu, and more.  The Prometheus exposition format is also being used as the basis for [OpenMetrics](https://opensource.googleblog.com/2018/08/openmetrics-project-accepted-into-cncf.html).

I think my error handling within the goroutine is suboptimal. I'd be happy to receive feedback on that or anything else. The configurable port 9101 was to just follow the port that Prometheus node_exporter uses (it's not set in stone by any means though).

When enabled, you should be able to connect to `http://localhost:9101` and see your metrics:

```
# HELP dcr_block_height Returns the latest block height
# TYPE dcr_block_height gauge
dcr_block_height 268862
# HELP dcr_peer_connections Returns a count of connections from peers
# TYPE dcr_peer_connections gauge
dcr_peer_connections 23
# HELP dcr_tickets_live Returns a count of live tickets
# TYPE dcr_tickets_live gauge
dcr_tickets_live 42
# HELP dcr_tickets_missed Returns a count of missed tickets
# TYPE dcr_tickets_missed gauge
dcr_tickets_missed 4
```
 I'll be writing a guide on how to set up some useful Prometheus alerts along with a [Grafana dashboard](https://twitter.com/stakeynet/status/1174820758339514368).

One useful metric gathered here is the block height metric. It's pulled from dcrwallet, so it ensures both dcrd and dcrwallet are working with the tip of the chain.  My testnet nodes are fairly fragile and this metric catches any of their failures including transient ones where processes are running but are actually hung.  On top of that the prometheus alertmanager can basically confirm that an error is happening a number of times before sending the actual alert, which is pretty useful for weeding out any short-term false positives.  My promql expression to create alert looks like this:

```yaml
- alert: TestnetBlockHeightStalled
        expr: dcr_block_height{group="testnet-voters"} < on() group_left max(dcr_block_height{group="testnet-voters"})
        for: 15m
        annotations:
          summary: Node is behind on best block height
```
And the other particularly useful metric is the live ticket count.  You can monitor whether all nodes agree.  When they don't, dcrstakepool becomes unresponsive. Live tickets are also the primary responsibility of the voting nodes, so it's pretty relevant whether they all agree about the current state of tickets.  I'll write some documentation about how to monitor these metrics, and how to address issues when those alarms sound off.

Note: cve-2019-16276 is relevant here as we're using net/http.  Patched Go versions are Go 1.13.1 and 1.12.10.


isuldor:
If the travis-ci error needs to be addressed, I may need some guidance as to how.

> backend/stakepoold/server.go:656:46: ctx.WalletConnection.GetStakeInfo undefined (type *rpcserver.Client has no field or method GetStakeInfo)

dajohi:
env GO111MODULE=on go mod tidy

commit go.mod and go.sum changes.

isuldor:
Okay, I reproduced the error locally after doing `go mod tidy` as well as ensuring that I was rebased on the latest master commit. It's weird that the only difference is a couple lines in go.mod referencing modern-go stuff?
```
-       github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
-       github.com/modern-go/reflect2 v1.0.1 // indirect
```

The error is `undefined method GetStakeInfo` so I presume I need to add it somewhere?  Not sure where though.

isuldor:
It passes now but the resulting build doesn't actually finish starting up.  Don't merge this for now.

isuldor:
Fixed travis-cl issue. This should be good to go now, pending any further comments or suggestions.

backend/stakepoold/config.go
```diff
@@ -80,6 +80,9 @@ type config struct {
 	RPCListeners     []string `long:"rpclisten" description:"Add an interface/port to listen for RPC connections (default port: 9113, testnet: 19113)"`
 	RPCCert          string   `long:"rpccert" description:"File containing the certificate file"`
 	RPCKey           string   `long:"rpckey" description:"File containing the certificate key"`
+	Prometheus       bool     `long:"prometheus" description:"Export metrics to be scraped by Prometheus"`
+	PrometheusWait   string   `long:"prometheuswait" description:"Amount of time wait between Prometheus metric updates (default: 1m)"`
```
JoeGruffins: This needs a small explanation of how to input the times. Also perhaps the type should be time.Duration. For comparison https://github.com/decred/dcrd/blob/29064448cb2e3a352ad0089b56a0b747725d0ac0/config.go#L108

jholdstock: Completely agree with this. I had no idea you could pass a `time.Duration` through a cli flag, thats really cool!

backend/stakepoold/server.go
```diff
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/http" // promhttp
```
jholdstock: > // promhttp

It isn't clear what this comment is trying to convey. Could you please either remove or add more detail?

backend/stakepoold/server.go
```diff
@@ -377,6 +382,27 @@ func runMain() error {
 		}()
 	}
 
+	if cfg.Prometheus {
+		var addr string
+		var wait time.Duration
+
+		if cfg.PrometheusListen == "" {
+			addr = ":9101"
```
jholdstock: Default values should be applied in `config.go`. For example...

https://github.com/decred/dcrstakepool/blob/625bcd23ec3901d7b732903b9255a1a03104e0d2/config.go#L40

https://github.com/decred/dcrstakepool/blob/625bcd23ec3901d7b732903b9255a1a03104e0d2/config.go#L327

This applies for both `PrometheusListen` and `PrometheusWait`

jholdstock:
Hey @isuldor, I added a few in-line comments, and also I agree with Joe that we should mention this feature in README. Maybe just a one-liner in README.md which links to a new document under `docs/`. This could include usage instructions and some sample Prometheus configs, like the one you posted above.

Also wanted to make you aware of https://github.com/lightninglabs/lndmon which already implements a Prometheus exporter for lnd. It should be possible to port this to dcrlnd **really** easily

backend/stakepoold/server.go
isuldor: sorry about that, this has been removed

backend/stakepoold/server.go
isuldor: Thanks for suggesting this!

backend/stakepoold/config.go
isuldor: Really good idea, thanks!

isuldor:
WIP
- [x] default values in config.go
- [ ] appcontext method
- [ ] shutdown channel
- [ ] readme one-liner

backend/stakepoold/server.go
```diff
@@ -595,3 +621,50 @@ func saveData(ctx *rpcserver.AppContext) {
...
+		Help: "Returns a count of missed tickets",
+	})
+	// Prometheus can be made to alert if nodes don't agree on block
+	// height for too long, indicative of a transient failure
+	promBlockHeight := promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "dcr_block_height",
+		Help: "Returns the latest block height",
+	})
+	// Periodically scrape for interesting statistics
+	go func() {
```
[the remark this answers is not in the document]
JoeGruffins: For example, if foo() is the thing you want to do periodically, this is the structure that I believe we have adopted:

someFunc(ctx context.Context, wg *sync.WaitGroup) {
    wg.Add(1)
    go func(){
        defer wg.Done()
    out:
        for {
            select {
                case <- ctx.Done():
                break out
                case <- time.After(WAITTIME):
                foo()
            }
        }
    }()
}

backend/stakepoold/config.go
```diff
@@ -264,6 +272,10 @@ func loadConfig() (*config, []string, error) {
 		PoolFees:   defaultPoolFees,
 		RPCKey:     defaultRPCKeyFile,
 		RPCCert:    defaultRPCCertFile,
+
+		Prometheus:       defaultPrometheus,
```
JoeGruffins: I don't think it's necessary to specify false since that is the zero value of a bool, but it's also fine if you want to specify it.  Just pointing that out, not asking for a change.

backend/stakepoold/stakepool/stakepool.go
```diff
@@ -472,6 +477,53 @@ func (spd *Stakepoold) UpdateUserDataFromMySQL() error {
...
+		Help: "Returns the latest block height",
+	})
+
+	go func() {
+		defer wg.Done()
+	out:
+		for {
+			select {
+			case <-ctx.Done():
+				break out
```
JoeGruffins: I asked you to do it this way, but jholdstock pointed out to me on my pr that we can just return here.

So, can you just return here and remove the "out" label above?

backend/stakepoold/stakepool/stakepool.go
```diff
@@ -472,6 +477,53 @@ func (spd *Stakepoold) UpdateUserDataFromMySQL() error {
...
+				if err == nil {
+					promConnectionCount.Set(float64(dcrdConnectionCount))
+				}
+				stakeInfo, err := spd.GetStakeInfo()
+				if err == nil {
+					promBlockHeight.Set(float64(stakeInfo.BlockHeight))
+					promTicketsLive.Set(float64(stakeInfo.Live))
+					promTicketsMissed.Set(float64(stakeInfo.Missed))
+				}
+				time.Sleep(wait)
```
JoeGruffins: It would be better if you made this into a case waiting for a tick on a channel from time.

Like it is now, the user will be forced to wait until this wait is over for the shutdown to happen.

If you do it like
for {
    case <-ctx.Done():
        return
    case <- time.After(wait):
        foo()
}
with no default case, then a shutdown signal will stop this loop even if in the middle of a wait.

jholdstock: agreed

backend/stakepoold/stakepool/stakepool.go
jholdstock: agreed

backend/stakepoold/stakepool/stakepool.go
isuldor: Thanks.  For some reason I was expecting the for loop to run almost continuously uninhibited, eating cpu cycles and sending a cpu core to 100%.  I'm running a build with this and it seems fine though?

JoeGruffins: Why doesn't or does?

without a default case the select should just wait and not use cpu.  What do you see?

isuldor: Oh!  I didn't realize the select was blocking if there is no default.  :mindblown:

jholdstock:
Code is looking good at this point. Just awaiting the doco updates

backend/stakepoold/server.go
```diff
@@ -361,6 +361,10 @@ func runMain(ctx context.Context) error {
 		}()
 	}
 
+	if cfg.Prometheus {
+		spd.ExportPrometheusMetrics(ctx, wg, cfg.PrometheusWait, cfg.PrometheusListen)
```
JoeGruffins: can you log an info here, something like "Serving Prometheus metrics at http://{ip}:{port}/metrics"

backend/stakepoold/stakepool/stakepool.go
```diff
@@ -472,6 +477,51 @@ func (spd *Stakepoold) UpdateUserDataFromMySQL() error {
...
+	})
+	promTicketsMissed := promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "dcr_tickets_missed",
+		Help: "Returns a count of missed tickets",
+	})
+	promBlockHeight := promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "dcr_block_height",
+		Help: "Returns the latest block height",
+	})
+
```
JoeGruffins: need to wg.Add(1) here

isuldor:
> Where did your changes to sample-stakepoold.conf go? we need those. Be sure to leave a small comment about how to input the times, same as the config.

I trampled them while I was fussing with git, I'll have to rewrite them lol

backend/stakepoold/stakepool/stakepool.go
isuldor: Yeesh, thanks. I had it in before but must have dropped it accidentally.

isuldor:
Ready for a final review.  Also, how would I squash these commits?  I've mostly just been cheating and rolling back HEAD, but I should probably learn to do it proper.

JoeGruffins:
I only ever use the command line, but I think what you need to do is something like,
git checkout master
git fetch upstream/master (note: you may not have upstream set as so)
git merge upstream/master

git checkout prometheus_instrumentation
git rebase -i master

And here delete all the comments that aren't yours
Save and quit

You will probably have merge conflicts, fix conflicts and add the files
git add foo.go
git rebase --continue

until
git status
reborts a clean status

then
git push --force

at any time during rebase, you can revert to before the rebase using
git rebase --abort

isuldor:
> The problem here is that you merged master onto your branch instead of rebasing your branch with master.
I see, maybe I'll just need to reopen the PR in a new branch.

JoeGruffins:
Ohhh. This looks awesome.  Give me a little time to test it out.

jholdstock:
Hey @isuldor,

The `/metrics` endpoint seems to work well for me, but I am having some issues when I try to shutdown stakepoold. Using CTRL+C I get

```
2019-10-26 12:28:44.435 [INF] STPK: Received signal (interrupt).  Shutting down...
```

and then the process just hangs indefinitely. I believe it is because the call to `http.ListenAndServe` is blocking and does not return. I'm chatting to @JoeGruffins about how best to resolve this, as he has been touching the shutdown signals recently.

JoeGruffins:
We are waiting for a shutdown and then shutting down a server here https://github.com/decred/dcrstakepool/blob/39945e3f78dbfe7cf55b0764addab22b4fcab615/server.go#L296
I think this before listening, and also listening in a go routine may be best

isuldor:
It now handles shutdown signals.  I have to push back on running the http server in a thread as I see no precedent for doing this (at least dcrstakepool doesn't appear to do that). And the Golang manual explains that the http server spins off each request into their own thread already, so it seems like overkill.

docs/prometheus-examples.md
```diff
@@ -0,0 +1,37 @@
...
+ 
+### Systemd service units
+These [service units](https://www.freedesktop.org/wiki/Software/systemd/) assume you've set up Prometheus under various `/opt` subdirectories, with binaries stored in `/opt/bin`.  They use an unprivileged user account to run the daemon. For your convenience, `ExecReload` has been configured so that you can reload configuration on the fly, e.g. 	`systemctl reload prometheus`.  Assuming you went through the quickstart already, here is how you could set up the `prometheus` daemon :
+```bash
+ sudo mkdir -p /opt/prometheus
+ sudo cp -r prometheus-2.13.1.linux-amd64/{consoles,console_libraries} /opt/prometheus/
+ sudo cp prometheus-2.13.1.linux-amd64/prometheus /opt/bin/
```
JoeGruffins: Starting on a clean ubuntu, I also had to create the /opt/bin directory.

Also the service requires the promtool binary as well. Also, the previously made prometheus.yml and alerting.rules.yml needs to be in the prometheus folder. If I was following your quickstart guide, then it's already in the user's folder, you could just change `sudo cp -r prometheus-2.13.1.linux-amd64/{consoles,console_libraries}` to have bash completion `*` at the end to simplify and take everything.

I agree that the person setting this up should be able to debug this, so it's up to you if this is intentionally left out for brevity.

docs/prometheus-examples.md
```diff
@@ -0,0 +1,37 @@
...
+ sudo systemctl daemon-reload
+ sudo systemctl enable prometheus
+ sudo systemctl start prometheus
+```
+
+### Blackbox Exporter
+[Blackbox exporter](https://github.com/prometheus/blackbox_exporter/) is a Prometheus daemon you can deploy to monitor http endpoints like `dcrstakepool` itself. See the relevant job named "health-checks" in `prometheus.yml`.  The service unit assumes configuration is in a directory `/opt/blackbox_exporter` owned by the `prometheus` user like in the example above.
+
+### Mobile push notifications
+The provided `alertmanager.yml` example sends alerts to the api of a service called [pushover.net](https://pushover.net/). For one time fee of $5, you can use their api to send push notifications directly to your Android or iOS mobile device.  Alertmanager supports numerous notification methods, including a generic webhook api, PagerDuty (a significantly more expensive service), OpsGenie (free for up to 5 users), and more.
```
JoeGruffins: Could you change "For a one time fee of $5" to something more ambiguous?  We can't promise that that fee will stay $5 or one time.  Maybe just "For a fee,"

isuldor:
I've changed up the error handling a bit so that it will `continue` on errors. ie. It should wait for the next interval period to try again. The two rpc methods I'm calling already print errors, so I opted to use `log.Debugf` so it won't print the same error twice when debugging is off.

isuldor:
@dajohi I believe this is good to go.

isuldor:
For what its worth, I've been running this in production for a few weeks now with a 30 second interval and have had zero missed votes in that time.

JoeGruffins:
@dajohi

PR #534, opened 2019-09-20
clean up LoadTemplates and link to explorer on stats page

chappjc:
Previously, `template.Must(httpTemplates, nil)` was pointless (`err=nil`).
This assigns the parsed templates directly to the `Templates` field.

Also define a template function map for more involved template tasks.
e.g. On the stats page, use the `netname` function for links. This could
also be done prior to executing the template, but it may be desirable
to have the actual network name with any number suffix (e.g. testnet3)
still available in the `Network` data field.

Add `debuglevel=debug` to the stakepoold and dcrstakepool config used
in harness.sh.

### DIFF
--- a/system/core.go
+++ b/system/core.go
@@ -62,6 +62,19 @@ func (application *Application) Init(APISecret, baseURL, cookieSecret string,
 	application.APISecret = APISecret
 }
 
+var funcMap = template.FuncMap{
+	"times": func(a, b float64) float64 {
+		return a * b
+	},
+	"netname": func(network string) string {
+		net := strings.ToLower(network)
+		if strings.HasPrefix(net, "testnet") {
+			return "testnet"
+		}
+		return net
+	},
+}
+
 func (application *Application) LoadTemplates(templatePath string) error {
 	var templates []string
 
@@ -82,17 +95,9 @@ func (application *Application) LoadTemplates(templatePath string) error {
 		return err
 	}
 
-	// Since template.Must panics with non-nil error, it is much more
-	// informative to pass the error to the caller (runMain) to log it and exit
-	// gracefully.
-	httpTemplates, err := template.ParseFiles(templates...)
-	if err != nil {
-		return err
-	}
-
-	application.Template = template.Must(httpTemplates, nil)
 	application.TemplatesPath = templatePath
-	return nil
+	application.Template, err = template.New("vsp").Funcs(funcMap).ParseFiles(templates...)
+	return err
 }
 
 func (application *Application) Close() {
--- a/views/stats.html
+++ b/views/stats.html
@@ -62,7 +62,7 @@
 					<div class="row">
 						<div class="col text-center bg-white py-2">
 							<p class="font-weight-bold text--size-13 mb-0">Ticket Price</p>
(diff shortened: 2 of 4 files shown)


system/core.go
```diff
@@ -62,6 +62,19 @@ func (application *Application) Init(APISecret, baseURL, cookieSecret string,
 	application.APISecret = APISecret
 }
 
+var funcMap = template.FuncMap{
+	"times": func(a, b float64) float64 {
+		return a * b
+	},
+	"netname": func(network string) string {
+		net := strings.ToLower(network)
+		if strings.HasPrefix(net, "testnet") {
```
jholdstock: #445 actually added this functionality already. 

https://github.com/decred/dcrstakepool/blob/625bcd23ec3901d7b732903b9255a1a03104e0d2/controllers/main.go#L170-L179

https://github.com/decred/dcrstakepool/blob/625bcd23ec3901d7b732903b9255a1a03104e0d2/controllers/main.go#L192

---

And we also have other places where the URL is being constructed in the template:

https://github.com/decred/dcrstakepool/blob/625bcd23ec3901d7b732903b9255a1a03104e0d2/views/tickets.html#L179

---

We should probably try and standardise all of this and re-use/refactor the existing code rather than adding new code. Perhaps we could construct the dcrdata URL once in `config.go` and just pass the full URL around, rather than constructing the URL in multiple places.

views/stats.html
```diff
@@ -62,15 +62,15 @@ <h1><span>Network Stats</span></h1>
 					<div class="row">
 						<div class="col text-center bg-white py-2">
 							<p class="font-weight-bold text--size-13 mb-0">Ticket Price</p>
-							<p class="mb-0 text--size-13">{{printf "%0.2f" .StakeInfo.Difficulty}}&nbsp;DCR</p>
+							<p class="mb-0 text--size-13"><a href="https://dcrdata.decred.org/charts?chart=ticket-price&zoom=month" target="_blank" rel="noopener noreferrer">{{printf "%0.2f" .StakeInfo.Difficulty}}&nbsp;DCR</a></p>
```
jholdstock: Hardcoded mainnet - deliberate?

jholdstock: Also I notice that we are using `dcrdata.decred.org` here, but the link added further down is `mainnet.decred.org` .

My understanding was that we should use `dcrdata.org`, but I could be wrong about that. I would like to try and keep consistency where possible though.

system/core.go
```diff
@@ -62,6 +62,19 @@ func (application *Application) Init(APISecret, baseURL, cookieSecret string,
 	application.APISecret = APISecret
 }
 
+var funcMap = template.FuncMap{
+	"times": func(a, b float64) float64 {
```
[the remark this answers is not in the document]
chappjc: Ah, in my fork of dcrstakepool, I was using times to convert missed as a fraction to a percent.  Can remove this, but I figured it was no big deal to keep as an example or utility.

system/core.go
chappjc: The reason for having hit here is that in some routes, `{{$.Network}}` is not passed through `(*MainController).getNetworkName` first.  On such routes, the "testnet3" is needed, so the number can't be stripped from it.  The solutions are: (1) add another data field passed to the template with both versions of the network name, or (2) strip it on the fly in template execution.

views/stats.html
chappjc: ah, no that's a mistake.  I think we've still using dcrdata.org since the two hosts (mainnet.dcrdata.org and testnet.dcrdata.org) exists, while the ones on decred.org still point to Insight.

PR #535, opened 2019-09-20
Define and use a custom HTTP request logger

chappjc:


system/middleware.go
```diff
@@ -108,3 +111,31 @@ func (application *Application) ApplyAuth(c *web.C, h http.Handler) http.Handler
...
+// Logger is a middleware that logs the start and end of each request, along
+// with some useful data about what was requested, what the response status was,
+// and how long it took to return. This should be used after the RequestID
+// middleware.
+func Logger(RealIPHeader string) func(c *web.C, h http.Handler) http.Handler {
+	return func(c *web.C, h http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			reqID := gojimw.GetReqID(*c)
+
+			log.Infof("[%s] Started %s %q, from %s", reqID, r.Method,
```
chappjc: This may be too verbose for some.  These can be made Debugf and/or the use of the request logging middleware can be made a config option.

PR #536, opened 2019-09-20
harness: add stakepoolcoldextkey to dcrwallet.conf

chappjc:


chappjc:
I was just reviewing the settings for the different processes.

PR #537, opened 2019-09-20
Add zipassets.sh to facilitate gzip_static use

chappjc:


chappjc:
Certainly.  I'll also add a comment in the sample-nginx.conf about this script to help utilize gzip_static.

PR #539, opened 2019-09-21
dcrstakepool: Add shutdown context

JoeGruffins:
Adds the file signal.go to handle interrupt signals.  On interrupt
safely shutdown server.  Also wait for sqlstore cleanup goroutine to
finish.


system/sqlstore.go
```diff
@@ -24,19 +26,26 @@ type SQLStore struct {
 
 // NewSQLStore returns a new SQLStore. The keyPairs are used in the same way as
 // the gorilla sessions CookieStore.
-func NewSQLStore(dbMap *gorp.DbMap, keyPairs ...[]byte) *SQLStore {
+func NewSQLStore(ctx context.Context, wg *sync.WaitGroup, dbMap *gorp.DbMap, keyPairs ...[]byte) *SQLStore {
```
jholdstock: I think `NewSQLStore` could follow the same pattern as introduced in #528. 

https://github.com/decred/dcrstakepool/blob/a2ca2e88a0dcb71214b43c75a48b573eae3898a1/backend/stakepoold/rpc/rpcserver/context.go#L797-L809

I think deferring the `wg.Done()` and using `return` instead of `break out` would be more consistent with the rest of the code, and it is a lot cleaner IMO.

signal.go
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2013-2014 The btcsuite developers
```
jholdstock: Rather than adding this file twice, are we able to reuse `signal.go` from #528?

server.go
```diff
@@ -5,13 +5,15 @@
...
+	"context"
 	"fmt"
 	"net"
 	"net/http"
 	"os"
 	"strings"
+	"sync"
 
-	"github.com/gorilla/context"
+	gorillactx "github.com/gorilla/context"
```
jholdstock: I'm not sure we should be using `gorilla/context`. [The repo](https://github.com/gorilla/context) is read-only and seems to be deprecated in favour of stdlib `context.Context`

signal.go
JoeGruffins: My thinking is that they may diverge at some point.
I guess that will work just as well tho.

signal.go
```diff
@@ -0,0 +1,51 @@
...
+// Use of this source code is governed by an ISC
```
JoeGruffins: Forgot the license too.

server.go
JoeGruffins: I can't find where this is doing anything, so I removed it.

In the entire code base it is only used here https://github.com/decred/dcrstakepool/blob/85435210c6a561f541e39e9c629ac2b67906764f/server.go#L183

It looks like the context stores its own data maps, but they will contain no data to clear unless it is previously set https://github.com/gorilla/context/blob/master/context.go

We are using gorilla csrf, but it says nothing about calling clear in its example https://github.com/gorilla/csrf/blob/7b29b05df5ae7dc777b92fab3dadee1b15d15ed6/csrf.go#L78-L126

JoeGruffins: And then I found this https://github.com/gorilla/sessions/issues/121
so I dunno.  I'll create an issue about it, since I'm not sure if this needed, or a problem.

system/sqlstore.go
jholdstock: You resolved this one but I dont think the change has been made?

JoeGruffins: I totally overlooked the "break out", thank you for finding and pointing that out.

Issue #532, opened 2019-09-19
eliminate all inline javascript


chappjc:
All JS should be coming from js files, not embedded in the html in `<script>` tags.  Part of the reason this is still done is because there are template variables doing into the inline scripts.  This should not be done like this.  Instead, the relevant html elements should have data tags and the JS (from .js files) should reference the html elements for this data.  e.g. The Stimulus.js pattern used in dcrdata is an example of how to do this formally.  See the stimulus "controllers" defined in https://github.com/decred/dcrdata/tree/master/public/js/controllers, and the corresponding `data-controller` and `data-*` attributes in the html templates in https://github.com/decred/dcrdata/tree/master/views.

jholdstock:
I will take this on

PR #540, opened 2019-09-23
Update to d3.js version 5.12.0

### LINKED ISSUE #533: update to modular d3 (and compatible d3pie)
The [d3.js](https://github.com/d3/d3/releases) being used is very old (v3.4.4 from Mar 25, 2014 from what I can determine).  Since then, there have  been two major releases.

D3 v4 ([4.0.0](https://github.com/d3/d3/releases/tag/v4.0.0) on Jun 28, 2016, and [4.13.0](https://github.com/d3/d3/releases/tag/v4.13.0) Jan 29, 2018) introduced a module design, that would allow only the required libraries (whatever is needed to make the pie charts) to be used.  This is desirable as the d3.min.js file pushing ~450 kB, which is ridiculously large for pie charts.

D3 v5 ([5.0.0](https://github.com/d3/d3/releases/tag/v5.0.0) on Jan 28, 2018, and [5.12.0](https://github.com/d3/d3/releases/tag/v5.12.0) on Sep 7, 2019).

D3 CHANGES.md: https://github.com/d3/d3/blob/master/CHANGES.md

[**d3pie.js**](https://github.com/benkeen/d3pie/releases) on the other hand has changed very little.  We seem to be using 0.1.8 or 0.1.9.  The latest, 0.2.1, supports d3 v4.  I have tested d3pie v0.2.1 with d3 v4.13.0 in dcrstakepool and it seems to work without issue.  I have not tried to identified the minimum set of d3 submodules, instead just using the monolithic version of d3 that (unnecessarily) includes everything.

In short, we should either drop d3/d3pi entirely and find a more efficient way of making pie charts, or we should strive to get D3 up-to-date (on v5) and using the minimum required modules, and d3pie working with D3 v5 if that means patching d3pie itself.

jholdstock:
- Chart legend is now drawn by d3 rather than manually in the html template
- Most of the javascript creating the charts is now in a dedicated file (good progress towards #532)
- Chart labels are now showing absolute values rather than percentages. This is because percentages are quite easy to estimate by looking at the size of the pie slices, but it is not possible to work out the absolute values.

Closes #533


jholdstock:
Requesting review from @chappjc

chappjc:
A big benefit of the newer d3 versions is the ability to pick and choose the modules you need.  You've opted for the default package, which includes everything, so that works but it's much bigger than needed.

See https://github.com/d3/d3/blob/master/API.md.
I think we'd just need d3-path, d3-selection, d3-scale (and it's deps except the time ones), and d3-shape.

jholdstock:
Done. Previously 1 file at `246kB`, now 9 files at `91kB`

PR #544, opened 2019-09-24
Simplify construction of dcrdata urls

jholdstock:
The dcrdata URL is now only built in one location rather than scattered
throughout the templates.

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -179,6 +179,12 @@ func (controller *MainController) getNetworkName() string {
 	return controller.Cfg.NetParams.Name
 }
 
+// DCRDataURL returns a URL for accessing dcrdata. It will automatically detect
+// mainnet or testnet depending on current config.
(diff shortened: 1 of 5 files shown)


controllers/main.go
```diff
@@ -1822,7 +1828,7 @@ func (controller *MainController) Tickets(c web.C, r *http.Request) (string, int
 	}
 
 	c.Env["IsTickets"] = true
-	c.Env["Network"] = controller.getNetworkName()
+	c.Env["DCRDataURL"] = controller.DCRDataURL()
```
chappjc: Really good approach for consistency, but it seems like doing the `fmt.Sprintf` repeatedly is unnecessary.  Perhaps `DCRDataURL` could be a `string` field of `MainController` that gets set on construction of the controller.

jholdstock: Nice idea. Implemented

PR #545, opened 2019-09-24
Deprecate unused Profile configs.

jholdstock:


itswisdomagain:
@jholdstock probably a good opportunity to completely remove previously deprecated config variables since 1.2.0 is released.

jholdstock:
Raised that in #546

Issue #512, opened 2019-09-03
a better rpcclient


JoeGruffins:
Currently we used the rpcclient located in dcrd.  It works fine except when we have a disconnect and autoreconnect is turned on.  While trying to reconnect, rpcclient accepts all new requests and waits for a reconnect, blocking, which creates issues as we are trying to show the results in real time.  To remedy this we have some easy fixes, but we really need to change the behaviour of rpcclient in order to get the benefits of rpcclient's autoreconnect.

JoeGruffins:
Food for thought from @jrick 

> The notifications were certainly the main issue I encountered with rpcclient in wallet. Applications really need to handle the disconnect and reconnect themselves in some sane way to ensure that state remains valid. Many issues begin to arise when you try to do it generically underneath them.
> 
> It's for a similar reason that I opposed adding round-robin support to rpcclient: your application may assume certain data exists based on prior calls, but that is no longer the case when you start querying a cluster. Load balancers can cause similar issues if your application isn't prepared for it, as I'm sure you are aware. And again back to the notifications, it would have broken all of the stateful ones. The network state and connections are best known and handled directly by the application.
> 
> This can be more work, but the benefit is large. For dcrwallet, I ended up making the wallet package the state of the application and implemented syncers on top of it which call mutator methods to extend the chain, add transactions, etc. If the client connection is lost, the syncer closes but the wallet state is left functional, ready to be synced by a new syncer later on. When it came time to add SPV support, this was possible by writing an alternative syncer instead of turning the wallet package into a spaghetti code disaster (more so than it already is). I can't say if this is the right approach for dcrstakepool, but it's an example of how an application can be better designed by handling the (very common) issue of network loss directly rather than abstracting it away.

JoeGruffins:
Lots of really good discussion here: https://github.com/decred/dcrd/issues/1877

JoeGruffins:
Maybe a good direction:
https://github.com/decred/dcrwallet/commit/f4b7458e5a72cf9f6cc2699162efa49a39bbb4d4#diff-f787792521e3d1834272d465f256cb3e

JoeGruffins:
Moar code:
https://github.com/decred/cspp/pull/18/files

JoeGruffins:
[...]

Issue #516, opened 2019-09-08
wallet conf: remove unneeded suggestion


JoeGruffins:
I think this can be removed now, correct? All rpc requests should now be coming from localhost.

https://github.com/decred/dcrstakepool/blob/a94b6966190e4c5e69af571fca030460c64b43f6/sample-dcrwallet.conf#L10-L11

JoeGruffins:
Or, maybe this isn't as simple as I thought, we still allow for dcrd and dcrwallet on a different machine than stakepoold. There isn't a matching comment for dcrd tho, and it is discouraged to use a different machine, so I still vote for its removal, as it is no longer necessary following our instructions.

jholdstock:
We don't really know what VSP operators are doing with their infrastructure. They may be using DNS, VLAN, they might use containers on a single host or on a cluster. I think it's best to leave this here as it could be beneficial for some users, and it doesn't really cause any harm.

JoeGruffins:
ok.

PR #552, opened 2019-09-30
[WIP] client/dcrwallet : grpc

JoeGruffins:
related to #512

This sets up up to start using the wallet's grpc.  The structure is almost exactly that which we are using for dcrstakepool <-> stakepoold. This pr only adds the functionality and a version check.

However, there are a few grpc methods missing that would let us do everything we do now.  Namely there doesn't seem to be a ListScripts. In the end this shouldn't be a problem, since I think it will end up that we need to "vote" without the wallet's help.

I did quite a bit of testing with pings and found the behaviour of disconnects and reconnects to behave as expected.  It is all done behind the scenes for us.

I'm still not sure this is the direction we need to go.  Just putting this out there and see what others think.

### DIFF
--- a/backend/stakepoold/config.go
+++ b/backend/stakepoold/config.go
@@ -209,6 +209,15 @@ func normalizeAddress(addr, defaultPort string) string {
 	return addr
 }
 
+// normalizeAddressReplacePort returns addr with the passed port
+func normalizeAddressReplacePort(addr, port string) string {
+	a, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		return addr
+	}
+	return net.JoinHostPort(a, port)
+}
+
 // normalizeAddresses returns a new slice with all the passed peer addresses
 // normalized with the given default port, and all duplicates removed.
 func normalizeAddresses(addrs []string, defaultPort string) []string {
--- a/backend/stakepoold/grpcserver.go
+++ b/backend/stakepoold/grpcserver.go
@@ -21,7 +21,7 @@ import (
 	xcontext "golang.org/x/net/context"
 
 	"github.com/decred/dcrd/certgen"
-	"github.com/decred/dcrstakepool/backend/stakepoold/rpc/rpcserver"
+	"github.com/decred/dcrstakepool/backend/stakepoold/rpc/server"
 	"github.com/decred/dcrstakepool/backend/stakepoold/stakepool"
 
 	"google.golang.org/grpc"
@@ -90,7 +90,7 @@ func interceptUnary(ctx xcontext.Context, req interface{}, info *grpc.UnaryServe
 	peer, peerOk := peer.FromContext(ctx)
 
 	// limit the time we take
-	ctx, cancel := context.WithTimeout(ctx, rpcserver.GRPCCommandTimeout)
+	ctx, cancel := context.WithTimeout(ctx, server.GRPCCommandTimeout)
 	// it is good practice to use the cancellation function even with a timeout
 	defer cancel()
 
@@ -191,7 +191,7 @@ func openRPCKeyPair() (tls.Certificate, error) {
 
 func startGRPCServers(stakepoold *stakepool.Stakepoold) (*grpc.Server, error) {
 	var (
-		server  *grpc.Server
+		svr     *grpc.Server
 		keyPair tls.Certificate
 		err     error
 	)
@@ -207,24 +207,24 @@ func startGRPCServers(stakepoold *stakepool.Stakepoold) (*grpc.Server, error) {
 		return nil, err
 	}
 	creds := credentials.NewServerTLSFromCert(&keyPair)
-	server = grpc.NewServer(grpc.Creds(creds), grpc.UnaryInterceptor(interceptUnary))
-	rpcserver.StartVersionService(server)
-	rpcserver.StartStakepooldService(stakepoold, server)
+	svr = grpc.NewServer(grpc.Creds(creds), grpc.UnaryInterceptor(interceptUnary))
+	server.StartVersionService(svr)
+	server.StartStakepooldService(stakepoold, svr)
 	for _, lis := range listeners {
 		lis := lis
 		go func() {
 			log.Infof("gRPC server listening on %s",
 				lis.Addr())
-			err := server.Serve(lis)
+			err := svr.Serve(lis)
 			log.Tracef("Finished serving gRPC: %v",
 				err)
 		}()
 	}
 
 	// Error when the server can't be started
-	if server == nil {
+	if svr == nil {
 		return nil, errors.New("gRPC service cannot be started")
 	}
 
-	return server, nil
+	return svr, nil
 }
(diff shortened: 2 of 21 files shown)


chappjc:
Well, I think that moving to gRPC instead of JSON-RPC is the wrong direction personally.  What's the motivation for that?

JoeGruffins:
@chappjc 
1. It seems to have solved connection issues without any extra code.  We also haven't had any problems with it as part of dcrstakepool <-> stakepoold.
2. It's endorsed by "g", newer, and has continued development. Implementing new methods also looks to be less complex.

It's fine if we don't use this, but is there as specific reason why rpc is better, or grpc is lacking?

jrick:
I think grpc is a better rpc protocol, but recently it hasn't received as much love and attention as the json-rpc server in dcrwallet.  This could mean there are some apis you want to use which aren't callable over grpc, but that can be fixed with additions.

If you need to use any kind of notifications, grpc is your only option. the dcrwallet json-rpc server does not produce any notifications.

chappjc:
>     1. It seems to have solved connection issues without any extra code.  We also haven't had any problems with it as part of dcrstakepool <-> stakepoold.

What does this mean "solved connection issues"?  What problem are you trying to solve by converting everything to gRPC?

> It's endorsed by "g", newer, and has continued development. Implementing new methods also looks to be less complex.

I don't count the above as positives.  JSON-RPC is exceedingly simple, so it is to be expected that continued development is not a requirement.  Conversely, I see gRPC as unnecessarily complex (at least for our purposes), with a good deal of moving pieces, which we count as unaudited dependencies, which actually generates unauditable Go code from yet another specification language (protobuf).

JoeGruffins:
@chappjc 
The "connection issues" are those caused by using dcrd/rpcclient. If autoreconnect is turned on a request will block forever by design. But we just want it to fail, so it was necessary to turn off autoreconnect and write our own. By contrast grpc has this behavior by default.

I'm not sure what you mean by the code is not auditable, it's not open source? Or we can't write tests? Do you think it would be prudent then to use rpc for the dcrstakepool<->stakepoold link as well? Why was that design chosen?

chappjc:
> The "connection issues" are those caused by using dcrd/rpcclient. If autoreconnect is turned on a request will block forever by design. But we just want it to fail, so it was necessary to turn off autoreconnect and write our own. By contrast grpc has this behavior by default.

Gotcha.  Although this PR is talking about a dcrwallet link so it wasn't clear if the issue was present with that too.  So gRPC does what when connection is down?  It autoreconnects, but it does not block forever when down?

> I'm not sure what you mean by the code is not auditable, it's not open source? Or we can't write tests? Do you think it would be prudent then to use rpc for the dcrstakepool<->stakepoold link as well? Why was that design chosen?

I mean, do we plan to review every commit made to  google.golang.org/grpc and https://godoc.org/github.com/golang/protobuf?  Also, take a look at the generated code the protoc-go spits out.

chappjc:
Not to say that these are absolute deal breaking issues with gRPC, but it's certainly not a no brainer to switch to it IMO.

chappjc:
As far as why the use of gRPC was chosen for dcrstakepool<->stakepoold... well, you'd have to ask the people who added that, but I wouldn't assume it was a solid decision.

JoeGruffins:
@chappjc 
Using a simple ping for testing, it seems that reconnects are happening as expected behind the scenes and we get an error when dcrwallet is down. "All comms in transient failure" or something like that.

So a
@jrick meh
@chappjc nah
@jholdstock ???
@itswisdomagain ???

chappjc:
Just trying to talk through this still.  What's the real problem **with the dcrwallet RPC connection** that this change is intended to solve?  The change comes at a real cost and with real risks, so there should be a good reason to match.  The chain state and notification issues aren't present with stakepoold->dcrwallet, right?

JoeGruffins:
@chappjc 
The real problem, I think, is just what I previously described.  To be honest, that autoreconnect fix works for me, as far as wallet commands are concerned, but it also seems overly complicated.  We want the client to just do that on its on.  I thought that implementing grpc would simplify the code.  That's all. And we are already using it elsewhere, so it's not adding a dependency. And if that's a problem then it's a bigger problem for dcrwallet, because funds are at stake.

It's true that there is only one stakepoold per wallet, but we don't know if dcrwallet is on the same machine as stakepoold.  Yes it should be but I made a pr about this and it was decided that, we just don't know.  #516 As you can input an ip, this makes sense. 

However, dcrd can't use grpc, so need to do the websocket stuff.  Maybe it would be less work for everyone to go through with the @davecgh suggestion

> I have been thinking for a while that rpcclient probably needs to be reworked in general to just provide the nice bidirectional jsonrpc to concrete types adapter code and leave the connection handling semantics to the consumer.

If that will make it more useful to all of us. It does seem a shame to write a bunch of code over and over...

chappjc:
OK, no more comments from me after this.   I'm just pointing out that the reconnect issues have only ever been a practical concern for the the connection to dcrd, which is _necessarily_ an over the internet in this case, not the wallet (even if an operator wants to make their life difficult by separating stakepoold and dcrwallet with uncontrolled hops on the internet, which by the way is only one extreme reason why rpclisten might need to be not localhost since these processes can be on the same LAN).  

Also I don't see it as overly complicated to have the dcrd reconnect fix (handle reconnect application-side rather than relying on rpcclient's autoreconnect feature) when the alternative is a big dependency that is effectively a black-box (even if it is already used in other places because now it's going to be in yet another spot).

Yes, it is valid to question whether it's a net positive feature for dcrwallet too.  Doesn't mean we should dive in and make the same leap of faith everywhere IMO.

chappjc:
> If that will make it more useful to all of us. It does seem a shame to write a bunch of code over and over...

I said I had my last comment, but I meant to respond to this bit by noting that the application (consumer) knows the best way to handle a lost and reestablished connection.  This is certainly the case when having to deal with chain state and possibly missed notifications from dcrd, although that is not the case in this PR ofc.  For stakepoold->dcrwallet, maybe your right and it's simple enough to just use gRPC.

JoeGruffins:
It would be simpler to just use the current wrapper for dcrwallet and call it a day... If that works.  Perhaps the reconnect back-off could be added. 
I'll wait for the other opinions tho.

jholdstock:
You can put me firmly in the "meh" camp - I see the benefits and drawbacks of both.

Each should be considered on a case-by-case basis. On a personal level I think I prefer gRPC but the overall direction of the project has been away from gRPC and toward JSON.

chappjc:
I've made my opinions about it clear, I think, but really this change isn't such a big deal.  It really may be net positive; I'm not convinced. It's unlikely to hurt vsp functionality.

It's also not me place.  No more push back from me.

JoeGruffins:
I'm going to start on the json-rpc version tomorrow, lol

JoeGruffins:
Thanks everyone who participated in the conversation, especially @chappjc !

PR #553, opened 2019-10-02
params: Remove unused WalletRPCServerPort

JoeGruffins:
I don't think we are using or plan to use the wallet rpc port anymore for dcrstakepool as all wallet communication will pass through stakepoold.


JoeGruffins:
@jholdstock 
I don't know what that is about.  I started using go 1.13 and it's doing that automatically when I build.

chappjc:
@JoeGruffins Same here.  I suppose those changes *could* go in another PR, but there's not much point in that IMO since the changes aren't affecting the build output at all.

PR #557, opened 2019-10-08
stakepoold: Add dcrwallet rpc client

JoeGruffins:
Adds a new dcrwallet rpc client.

The end effect is almost exactly what we have now. The plus is that we have full control over the connection, and the code is much simpler.  It's only what we need.  The connection code is completely separate from the rpc calls.  The calls can also be used by another application trying to accomplish the same effect, hence the new module.  It also set's up nicely to change the dcrd client over to a similar, manageable solution.


backend/stakepoold/stakepool/stakepool.go
```diff
@@ -333,13 +341,13 @@ func (spd *Stakepoold) AddMissingTicket(ticketHash []byte) error {
 		return err
 	}
 
-	tx, err := spd.WalletConnection.RPCClient().GetRawTransaction(hash)
+	tx, err := spd.NodeConnection.GetRawTransaction(hash)
```
JoeGruffins: As far as I can tell this has been a dcrd rpc from the beginning...

JoeGruffins: dcrd's command
https://github.com/decred/dcrd/blob/ee4a0e2e2a0e7dc2e71ae807e7586c5fe03ed407/rpc/jsonrpc/types/chainsvrcmds.go#L682-L689

dcrwallets's
https://github.com/decred/dcrwallet/blob/fe6e4c8c0cce1065111cacd5b11193595b7e386f/rpc/jsonrpc/types/methods.go#L489-L493

JoeGruffins: So, perhaps this was what we wanted, but the naming was just wrong, let me look into what it's used for a bit more.

JoeGruffins: It's used for the wallet rpc "addticket", which seems to accept a raw has based on the variable name here, but I'm unable to find a comment that confirms this.

https://github.com/decred/dcrd/blob/eb5bb6d4192acbf57dfedcf57aa6654adae10a6d/rpcclient/wallet.go#L2325-L2341

JoeGruffins: I order to use getrawtransaction on dcrd "The transaction index must be enabled to query the blockchain (specify --txindex)"

JoeGruffins: for transaction:
cc9d15715a51a0981a6a1911863122279fb05a1b2ff12f067762fc3317e85300

dcrwallet's hex:
0100000002380b3eadc0e9ac8446e350d374fb33e626e808f1f57d0f3f58adf2d7c663c1af0000000000ffffffff380b3eadc0e9ac8446e350d374fb33e626e808f1f57d0f3f58adf2d7c663c1af0100000000ffffffff05c3224c8c01000000000018baa914515d6c01df8f3d908da72da81204b51b26a864da8700000000000000000000206a1edc29731dd5919af414329223d7f64448c17efa95dcd60300000000000058000000000000000000001abd76a914000000000000000000000000000000000000000088ac00000000000000000000206a1e79d29020d3a558bf08ed3a7512f5ff6c992b77561361488c010000000058000000000000000000001abd76a914000000000000000000000000000000000000000088ac000000000000000002dcd6030000000000**00000000ffffffff**6b483045022100874e26f9ef3a28d5f0e1e72da61fca07cb35df019507bb124c7a33389cf8a55d02205c87d0746730bbb0f8b0e1d9b8e703bfbe4b69adf5dcd32cfa303ddfe31e2adc0121039ff5a6a373eb15d9b71505e78a9d39750b1e9616bc4af7fef82ee32cce9dc20b1361488c01000000**00000000ffffffff**6a47304402206f6a934e4206730d0d43c78ffcea9ca05c56846f751baac34e9e256d1558d1fb0220413a9ab43446c39bdf37fa7ddf300f5f7bbd138fd4d4ef60ff3f6f0be2e8910b0121039ff5a6a373eb15d9b71505e78a9d39750b1e9616bc4af7fef82ee32cce9dc20b

dcrd's hex:
0100000002380b3eadc0e9ac8446e350d374fb33e626e808f1f57d0f3f58adf2d7c663c1af0000000000ffffffff380b3eadc0e9ac8446e350d374fb33e626e808f1f57d0f3f58adf2d7c663c1af0100000000ffffffff05c3224c8c01000000000018baa914515d6c01df8f3d908da72da81204b51b26a864da8700000000000000000000206a1edc29731dd5919af414329223d7f64448c17efa95dcd60300000000000058000000000000000000001abd76a914000000000000000000000000000000000000000088ac00000000000000000000206a1e79d29020d3a558bf08ed3a7512f5ff6c992b77561361488c010000000058000000000000000000001abd76a914000000000000000000000000000000000000000088ac000000000000000002dcd6030000000000**de62030008000000**6b483045022100874e26f9ef3a28d5f0e1e72da61fca07cb35df019507bb124c7a33389cf8a55d02205c87d0746730bbb0f8b0e1d9b8e703bfbe4b69adf5dcd32cfa303ddfe31e2adc0121039ff5a6a373eb15d9b71505e78a9d39750b1e9616bc4af7fef82ee32cce9dc20b1361488c01000000**de62030008000000**6a47304402206f6a934e4206730d0d43c78ffcea9ca05c56846f751baac34e9e256d1558d1fb0220413a9ab43446c39bdf37fa7ddf300f5f7bbd138fd4d4ef60ff3f6f0be2e8910b0121039ff5a6a373eb15d9b71505e78a9d39750b1e9616bc4af7fef82ee32cce9dc20b

You can see only a slight difference in the output hexes...

JoeGruffins: after quite a bit of sleuthing

<details close>
<summary> breakdown</summary>

0100000002380b3eadc0e9ac8446e350d374fb33e626e808f1f57d0f3f58adf2d7c663c1af0000000000ffffffff380b3eadc0e9ac8446e350d374fb33e626e808f1f57d0f3f58adf2d7c663c1af0100000000ffffffff05c3224c8c01000000000018

vout 0 [baa914515d6c01df8f3d908da72da81204b51b26a864da87]

0000000000000000000020

vout 1 [6a1edc29731dd5919af414329223d7f64448c17efa95dcd60300000000000058]

000000000000000000001a

vout 2 [bd76a914000000000000000000000000000000000000000088ac]

0000000000000000000020

vout 3 [6a1e79d29020d3a558bf08ed3a7512f5ff6c992b77561361488c010000000058]

000000000000000000001a

vout 4 [bd76a914000000000000000000000000000000000000000088ac]

000000000000000002

dcd6030000000000 

block height [de620300] 

block index [08000000] 

6b

vin 0 [483045022100874e26f9ef3a28d5f0e1e72da61fca07cb35df019507bb124c7a33389cf8a55d02205c87d0746730bbb0f8b0e1d9b8e703bfbe4b69adf5dcd32cfa303ddfe31e2adc0121039ff5a6a373eb15d9b71505e78a9d39750b1e9616bc4af7fef82ee32cce9dc20b]

1361488c01000000de62030008000000 6a

vin 1 [47304402206f6a934e4206730d0d43c78ffcea9ca05c56846f751baac34e9e256d1558d1fb0220413a9ab43446c39bdf37fa7ddf300f5f7bbd138fd4d4ef60ff3f6f0be2e8910b0121039ff5a6a373eb15d9b71505e78a9d39750b1e9616bc4af7fef82ee32cce9dc20b]


</details>

the difference translates into a null block height and index for dcrwallet.

Not sure if the height and index are required yet.

JoeGruffins: Another difference is that the wallet only knows about tx that concern it.

JoeGruffins: So, thinking about the last reason, I guess this does need to be a dcrd command, and --txindex must be specified.  If I understand the purpose for this command correctly, the wallet must not have seen that txid, so we need to ask dcrd...  I think this change is correct.

backend/stakepoold/rpcclient.go
```diff
@@ -78,27 +81,22 @@ func connectWalletRPC(ctx context.Context, wg *sync.WaitGroup, cfg *config) (*st
...
 
-	ntfnHandlers := getWalletNtfnHandlers()
-
-	// New also starts an autoreconnect function.
-	dcrwClient, err := stakepool.NewClient(ctx, wg, connCfgWallet, ntfnHandlers)
+	conn, err := client.NewConn(ctx, wg, testopts)
 	if err != nil {
-		log.Errorf("Verify that username and password is correct and that "+
-			"rpc.cert is for your wallet: %v", cfg.WalletCert)
+		log.Errorf("new syncer error: %v", err)
```
JoeGruffins: This should say new client error. Will change after a review.

backend/stakepoold/ntfnhandlers.go
```diff
@@ -30,9 +28,8 @@ func getNodeNtfnHandlers(spd *stakepool.Stakepoold) *rpcclient.NotificationHandl
 				BlockHeight: blockHeight,
 				SmTickets:   ticketsFixed,
 			}
-			// Wait for a wallet connection if not connected.
-			<-spd.WalletConnection.Connected()
-			spd.SpentmissedTicketsChan <- smt
+			// Wait for a wallet connection if not connected or shutdown.
```
JoeGruffins: ~or shutdown~

JoeGruffins:
adding rpc method tests, give me a  few days

JoeGruffins:
Wallet calls now have tests that prove for the given variables they produce the expected JSON and the expected response is parsed correctly. All rpc commands have been tested individually, and on my test site on testnet. Users created and tickets bought without problems.

JoeGruffins:
@jholdstock review

JoeGruffins:
The dcrd sample configuration file has be moved to its own pr.

To give a little more explanation about where the code in the pr comes from.

connection.go:
Ths was our client.go in the stakepool package.  It sets up a new connection through gorilla ws client via wsrpc and also satifies the Caller interface specified by the dcrwallet/calls.go package. This allows it to hold any arbitrary calls and pass JSON RPC commands while keeping the connection alive with an autoreconnect function.  The flow of the autoreconnect function comes from here https://github.com/decred/dcrd/blob/5e1fbdf54eb92cd0e0bd0a670dd7c6aa0d288e91/rpcclient/infrastructure.go#L663-L731 with the added benefit of being able to break out in between retries.  It is also different from our previous client.go in that I have simplified waiting for a connection to perform some action using the client.TODO(func()) function.

calls.go
This file defines the Caller interface used to perform all JSON RPC calls and is an idea taken from dcrwallet https://github.com/decred/dcrwallet/blob/master/rpc/client/dcrd/calls.go Using the same pattern means that we can use calls defined there, or anywhere else.  It is a clean way to abstract the calls from connection and syncing logic.  The comments and logic for all calls comes from the dcrd/rpcclient repo.  I have removed some (I think) redundant, obvious inline comments like "converting so and so to string". At this stage, it is also much easier to read our calls, as the dcrd/rpcclient calls are spread all over the place. That ambiguity has lead to a mistake that is outlined above, where a call is made on the wrong client and not been noticed for a long time...

calls_test.go
This is an original test.  I think it proves that for the expected input we get the expected output.  It tests the parsing abilities of the calls.  This is easy to do thanks to the Caller interface.  I would dare say this makes ours the best tested rpc client package in all of the decred.org repo.  Feel free to correct me there.  Except for the second version test, all tests are actual data taken from my testnet and mainnet nodes.

jholdstock:
Thanks a lot for making those changes and providing the extra detail, that will help a lot with reviewing (which I am about to begin).

I notice there's a conflict which needs resolving before this can be merged.

JoeGruffins:
@jholdstock 
conflicts ok.
Also, I refactored AddMissingTicket. The correct way to proceed there can be decided in #569 
Also, also, the TODO(func) can be made internal in #563 when notifications are also moved into the client package. It is lower case there now.

JoeGruffins:
Judging from matrix comments, it looks like rpcclient may be in trouble. Would be nice to get this in and not have to worry about that dependency...

JoeGruffins:
After careful consideration, I think it is in the best interest of dcrstakepool not to export this as a separate module. It only creates extra work for us and complicates things. It is also much easier to turn it into a module at some point in the future than to have someone start using it and decide not to export it after the fact. Anyhow, it does not need to be in this pr.

PR #560, opened 2019-10-14
Remove all in-line JavaScript

### LINKED ISSUE #532: eliminate all inline javascript

jholdstock:
Closes #532

### DIFF
--- a/public/js/admintickets.js
+++ b/public/js/admintickets.js
@@ -1,0 +1,13 @@
+(function () {
+    // Change checkbox state if user clicks anywhere on the row.
+    // Not just the actual checkbox
+    var rows = document.getElementsByTagName("tr")
+    for (var i = 0; i < rows.length; i++) {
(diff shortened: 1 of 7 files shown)


chappjc:
Nice!  I'm going to test this out.

jholdstock:
Well spotted - this actually affected all of the charts. Fixed now

PR #563, opened 2019-10-15
[WIP] stakepoold: Add dcrd RPC client

### LINKED ISSUE #512: a better rpcclient

JoeGruffins:
Work on top of #557 
closes #512 

Using the same patter as the new dcrwallet client, applied to dcrd.  Also registers for notifications.

This changes behavior in a small way.  rpcclient would cache all commands if there was a disconnect, but now we will error immediately, just as dcrwallet does.

This pr's purpose is not to handle chainstate better, but to set us up to handle chainstate better, by giving us a highly customizable rpc client.  From here the syncing mechanism can be whatever works best for us, and can be straightforward in its implementation.


JoeGruffins:
AAAAAnd forgot to make it a draft.

backend/stakepoold/server.go
```diff
@@ -310,23 +307,17 @@ func runMain(ctx context.Context) error {
 		log.Infof("block %v hash %v came in during GetTickets, refreshing...",
 			afterHeight, afterHash)
 	}
-
-	if err = nodeConn.NotifyBlocks(); err != nil {
-		fmt.Printf("Failed to register daemon RPC client for "+
-			"block notifications: %s\n", err.Error())
-		return err
-	}
```
JoeGruffins: We were registering for blocks notifications but ignoring the notification. Is there a reason for this that I am not seeing? I think this will be used in the future though.

jholdstock: Interesting observation. It was added in #178, I'm not sure why. Perhaps @chappjc remembers?

Issue #564, opened 2019-10-15
Improve golint score


JoeGruffins:
As seen here https://goreportcard.com/report/github.com/decred/dcrstakepool our golint score is only 63% and could easily be improved.  It is mostly due to lack of comments on exported functions.

chappjc:
Just wanted to interject that if you guys are gunning for a 2.0 at some point in the near future, you may want to consider unexporting stuff that has no conceivable external use.

jholdstock:
The last refresh on the report card was over a year ago. I just refreshed it and we are now at 64% :stuck_out_tongue:

JoeGruffins:
Going to try to knock out all the golint issues today.

PR #566, opened 2019-10-22
Report card

jholdstock:
I had some spare time, so here's a little work towards #564


backend/stakepoold/server.go
```diff
@@ -459,60 +460,55 @@ func loadData(spd *stakepool.Stakepoold, dataKind string) error {
...
 
-	// shouldn't get here -- data dir is created on startup
-	return errors.New("loadData - path " + spd.DataPath + " does not exist")
+	fullPath := filepath.Join(spd.DataPath, lastseen)
+
+	r, err := os.Open(fullPath)
+	if err != nil {
+		return err
+	}
+
```
JoeGruffins: I think that r (*os.File) should be closed at some point, although the godoc doesn't explicitly say so.

jholdstock: Done

backend/stakepoold/server.go
```diff
@@ -459,60 +460,55 @@ func loadData(spd *stakepool.Stakepoold, dataKind string) error {
 		return errors.New("unhandled data kind of " + dataKind)
 	}
 
-	if fileExists(spd.DataPath) {
-		files, err := ioutil.ReadDir(spd.DataPath)
-		if err != nil {
-			return err
-		}
+	if !fileExists(spd.DataPath) {
+		return errors.New("loadData - path " + spd.DataPath + " does not exist")
```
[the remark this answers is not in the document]
jholdstock: Done. Avoiding string concat is consistent with the rest of the codebase.

I'm not completely sure, but I think it's slightly more performant. As I understand, `"a" + "b" + "c"` results in three string allocations whereas `fmt.Sprintf("%s%s%s", a, b , c)` requires only one.

PR #567, opened 2019-10-22
multi: drop udb usage

dajohi:
@jrick says udb should be internal only

### DIFF
--- a/backend/stakepoold/server.go
+++ b/backend/stakepoold/server.go
@@ -26,7 +26,6 @@ import (
 	"github.com/decred/dcrstakepool/helpers"
 	"github.com/decred/dcrstakepool/signal"
 	"github.com/decred/dcrwallet/wallet/v3/txrules"
-	"github.com/decred/dcrwallet/wallet/v3/udb"
 
 	// register database driver
 	_ "github.com/go-sql-driver/mysql"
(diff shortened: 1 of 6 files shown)


backend/stakepoold/server.go
```diff
@@ -77,7 +76,7 @@ func calculateFeeAddresses(xpubStr string, params *chaincfg.Params) (map[string]
 	}
 
 	// Derive from external branch
```
JoeGruffins: Since we lose this information without the name, could you either include this comment in all affected places, or remake the constant in helpers package?

jholdstock: Agreed, a constant instead of magic numbers would be nice please.

JoeGruffins:
Looks like an import is missing.

helpers/address.go
```diff
@@ -11,6 +11,12 @@ import (
 	"github.com/decred/dcrd/hdkeychain/v2"
 )
 
+const (
+	// ExternalBranch is a helper value that needs to
+	// match dcrwallet's udb.ExternalBranch
+	ExternalBranch = 0
```
JoeGruffins: Could you also keep the type, uint32?
https://github.com/decred/dcrwallet/blob/d273b5687adfac4e74ba3fa21370da31158b1580/wallet/udb/addressmanager.go#L76

jholdstock: I'm curious why this should matter?

JoeGruffins: Because that's what it is. And the comment above says it needs to match...

PR #569, opened 2019-10-24
dcrd.conf: Add dcrd sample config file

JoeGruffins:
The purpose for turning on txindex is outlined in this pr #557 
I think it will be best to actively check that this is turned on on startup as well, if others agree.  I will make a seperate issue for that.

Otherwise this sample config follows the patterns set down by the the other sample configuration files.

### DIFF
--- a/backend/stakepoold/stakepool/stakepool.go
+++ b/backend/stakepoold/stakepool/stakepool.go
@@ -333,7 +333,7 @@ func (spd *Stakepoold) AddMissingTicket(ticketHash []byte) error {
 		return err
 	}
 
-	tx, err := spd.WalletConnection.RPCClient().GetRawTransaction(hash)
+	tx, err := spd.NodeConnection.GetRawTransaction(hash)
 	if err != nil {
 		log.Errorf("AddMissingTicket: GetRawTransaction rpc failed: %v", err)
 		return err
--- a/sample-dcrd.conf
+++ b/sample-dcrd.conf
@@ -1,0 +1,12 @@
+; Debug is very useful to see more activity.
+debuglevel=debug
+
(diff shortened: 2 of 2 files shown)


jholdstock:
Copying your comment from #557:

> So, thinking about that last difference I guess this does need to be a dcrd command, and --txindex must be specified. If I understand the purpose for this command correctly, the wallet must not have seen that txid, so we need to ask dcrd... I think this change is correct.
> 
> This command being AddMissingTicket... That's where the getrawtransaction and addticket rpc are located.

If I understand correctly, the motivation for this change to use dcrd because it will know any **any** transaction, rather than just ones relevant to the current wallet. Is this actually necessary? The VSP should only be dealing with transactions related to its own voting wallets afaik

JoeGruffins:
There is a lack of comments in this area as to why there would be missing tickets in the first place.

The only thing I can imagine is that a wallet was disconnected when we tried to import the redeem script. In which case the transaction would not be in the wallet. However it should show up after importing and rescanning the script.  Without the script it wouldn't really matter anyway tho.

But really the motivation is to match the intent of the original author, because calling GetRawTransaction on dcrwallet is a mistake. It must be called on dcrd, or changed to GetTransaction. The current code can only error. It doesn't work.

JoeGruffins:
It looks like someone hit this. The reason was an out of sync dcrd/dcrwallet.

Also wrote some code that shows that this does work as-is as long ad txindex=1 is set in dcrd:  https://github.com/JoeGruffins/pyscripts/blob/master/testrpc/testrpc.go

Although called on wallet, the command is routed to dcrd... apparently. I guess it doesn't really matter where we send the call currently, however I would prefer this be the node if that's where it is being routed to.

PR #575, opened 2019-12-15
readme: Simplify and update information

### LINKED ISSUE #508: README suggestions
Quoting @JoeGruffins in #505

> I think the current instructions are written:git clone (no help), build + copy, update, prerequisites, configure + copy, run(with a build thrown in)
> Maybe this could be made a little cleaner. Looking at https://github.com/decred/dcrd/blob/master/README.md
> 
> The installation section is labeled "Installation and Updating", maybe we could do the same.
> 
> {{YOUR_GO_MODULE_PATH}} is mentioned and then abandoned. Is it really necessary to mention? dcrd gets the same information across in just a few lines, with windows instructions. If the operator already installed dcrd and can run it they shouldn't be having a problem building. Anyway, I think build instructions can be shortened and maybe the mention to copying should be in a different place.
> 
> And maybe the updating section isn't necessary. dcrd uses the above section for both. There is already an upgrade guide in the release notes.
> 
> As mentioned above, I think copying the binaries instructions can be moved with the other copy commands, or under there respective sections, and doesn't belong in build.
> 
> I think the build command at the end for dcrstakepool should be removed as it should already be built. Also, if our instructions are to build and then copy it to the server, it won't be under the $GOPATH, there will probably be no $GOPATH on their frontend server if they were following our setup instructions, build on a different machine, and copied to the webserver. And then the next line says that it must be in that folder. So is the correct procedure to build everything on the frontend and export from there? If so the setup should be written so.
> 
> Also, we don't have a Documentation section with a link to docs. dcrd does and I think it's a good idea.

JoeGruffins:
closes #508 

Simplify installation instructions. Remove components section. Be more
explicit with setup instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ vote on their behalf when the ticket is selected.
 was approved by stakeholders to rename "Stakepool" to "Voting Service Provider", a.k.a. "VSP".
 These names are used interchangably in this repository.
 
+
 ## Architecture
 
 ![Voting Service Architecture](docs/img/architecture.png)
@@ -26,6 +27,7 @@ These names are used interchangably in this repository.
 - The architecture is subject to change in the future to lessen the dependence
   on MySQL.
 
+
 ## Test Harness
 
 A test harness is provided in `./harness.sh`. The test harness uses tmux to start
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -146,32 +116,43 @@ $ dcrctl --wallet accountsyncaddressindex teststakepoolfees 0 10000
 
 - Create the wallets.  All wallets should have the same seed.  **Backup the seed
   for disaster recovery!**
+- Log into wallet servers separatly and create wallets one at a time using the
```
jholdstock: separately*

README.md
```diff
@@ -26,81 +27,48 @@ These names are used interchangably in this repository.
...
+- **Git**
 
-```bash
-$ cd {{YOUR_GO_MODULE_PATH}}/github.com/decred/dcrstakepool
-$ go build
-```
+  Installation instructions can be found at https://git-scm.com
 
-#### Backend - stakepoold
+To build and install from a checked-out repo, run `go install . ./backend/stakepoold/...`
```
jholdstock: We dont need `...` here as there is no further dirs to work through. `go install . ./backend/stakepoold` is enough

JoeGruffins: I was thinking the dots pulled dependencies, but maybe I'm thinking of go get, and maybe that's not even an issue with modules?

README.md
```diff
@@ -26,81 +27,48 @@ These names are used interchangably in this repository.
...
-`$GOPATH`, the following instructions use the placeholder
-`{{YOUR_GO_MODULE_PATH}}` to refer to wherever you checkout your Go code.
+- **Go 1.11 ~ 1.13**
 
-#### Frontend - dcrstakepool
+  Installation instructions can be found here: https://golang.org/doc/install.
+  It is recommended to add `$GOPATH/bin` to your `PATH` at this point.
 
-Build dcrstakepool and copy it to the web server.
+- **Git**
```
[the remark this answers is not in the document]
JoeGruffins: It was in the dcrd readme, but probably nobody needs that info at this point, I'll take it out.

README.md
jholdstock: `...` is essentially for recursion down into the dir tree.

The command `go help packages` provides a nice explanation

PR #577, opened 2019-12-18
multi: Comment all exported functions.

### LINKED ISSUE #564: Improve golint score

JoeGruffins:
Closes #564 

Unexport one method. Change all occurrences of Id as a field with ID.
 
Changing the Id fields to ID isn't actually necessary to get the score down. So let me know if anyone disagrees with that.  I think it is more uniformed this way. However, the fields in the db are already mostly "somethingId" so...

https://goreportcard.com/report/github.com/JoeGruffins/dcrstakepool

I haven't touched the gocyclo. If that's something that should be worried about we can make a new issue.

### DIFF
--- a/backend/stakepoold/stakepool/stakepool.go
+++ b/backend/stakepoold/stakepool/stakepool.go
@@ -34,7 +34,7 @@ var (
 	ticketTypeSpentMissed = "SpentMissed"
 )
 
-// Stakepoold is stores everything related to stakepoold.
+// Stakepoold stores everything related to stakepoold.
 type Stakepoold struct {
 	sync.RWMutex
 
--- a/backend/stakepoold/stakepool/stakepool.go
+++ b/backend/stakepoold/stakepool/stakepool.go
@@ -34,6 +34,7 @@ var (
 	ticketTypeSpentMissed = "SpentMissed"
 )
 
+// Stakepoold is stores everything related to stakepoold.
 type Stakepoold struct {
 	sync.RWMutex
 
@@ -59,12 +60,14 @@ type Stakepoold struct {
 	Testing                bool // enabled only for testing
 }
 
+// NewTicketsForBlock stores tickets from a NewTickets notification.
 type NewTicketsForBlock struct {
 	BlockHash   *chainhash.Hash
 	BlockHeight int64
 	NewTickets  []*chainhash.Hash
 }
 
+// SpentMissedTicketsForBlock stores tickets from a SpentMissedTickets notification.
 type SpentMissedTicketsForBlock struct {
 	BlockHash   *chainhash.Hash
 	BlockHeight int64
@@ -78,6 +81,7 @@ type VotingConfig struct {
 	VoteBitsExtended string
 }
 
+// WinningTicketsForBlock stores tickets from a WinningTickets notification.
 type WinningTicketsForBlock struct {
 	BlockHash      *chainhash.Hash
 	BlockHeight    int64
@@ -232,6 +236,7 @@ func (spd *Stakepoold) getticket(wg *sync.WaitGroup, nt *ticketMetadata) {
 		strings.ToLower(nt.ticketType), nt.ticket)
 }
 
+// UpdateTicketData moves ignored low fee tickets to live tickets for voting.
 func (spd *Stakepoold) UpdateTicketData(newAddedLowFeeTicketsMSA map[chainhash.Hash]string) {
 	spd.Lock()
 
@@ -266,6 +271,8 @@ func (spd *Stakepoold) UpdateTicketData(newAddedLowFeeTicketsMSA map[chainhash.H
 	}()
 }
 
+// UpdateTicketDataFromMySQL moves ignored low fee tickets to live tickets for
+// voting based on information pulled from the DB.
 func (spd *Stakepoold) UpdateTicketDataFromMySQL() error {
 	start := time.Now()
 	newAddedLowFeeTicketsMSA, err := spd.UserData.MySQLFetchAddedLowFeeTickets()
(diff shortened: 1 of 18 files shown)


backend/stakepoold/stakepool/stakepool.go
```diff
@@ -34,6 +34,7 @@ var (
 	ticketTypeSpentMissed = "SpentMissed"
 )
 
+// Stakepoold is stores everything related to stakepoold.
```
jholdstock: remove `is`

controllers/main.go
```diff
@@ -49,6 +49,7 @@ const (
 	agendasCacheLife = time.Hour
 )
 
+// Config holds all the data used to create an new MainController.
```
jholdstock: a new*

system/sqlstore.go
```diff
@@ -71,7 +71,7 @@ func (s *SQLStore) New(r *http.Request, name string) (*sessions.Session, error)
 	err = securecookie.DecodeMulti(name, c.Value, &session.ID, s.codecs...)
 	if err != nil {
 		// these are not the sessions you are looking for
-		log.Infof("sqlstore: New: unable to decode cookie: %v", err)
+		log.Debugf("sqlstore: New: unable to decode cookie: %v", err)
```
jholdstock: Why is the log level changing here?

system/core.go
```diff
@@ -97,10 +101,13 @@ func (application *Application) LoadTemplates(templatePath string) error {
 	return err
 }
 
+// Close prints "Application.Close() called"
```
jholdstock: Wtf is this func... this is worth investigating, because this is very strange. Don't do it in this PR though, this one is already uncomfortably big.

helpers/auth.go
```diff
@@ -30,6 +32,8 @@ func EmailChangeComplete(dbMap *gorp.DbMap, token models.UserToken) error {
 	return err
 }
 
+// EmailChangeTokenExists checks whether the token exits and returns the
```
jholdstock: exists*

helpers/auth.go
```diff
@@ -51,6 +56,8 @@ func EmailExists(dbMap *gorp.DbMap, email string) (*models.User, error) {
 	return &user, err
 }
 
+// EmailVerificationTokenExists checks whether the token exits and returns User
```
jholdstock: exists*

helpers/auth.go
```diff
@@ -62,18 +69,23 @@ func EmailVerificationTokenExists(dbMap *gorp.DbMap, token models.UserToken) (*m
...
 
+// EmailVerificationComplete Changes a users EmailVerified value to 1.
 func EmailVerificationComplete(dbMap *gorp.DbMap, token models.UserToken) error {
 	_, err := dbMap.Exec(`UPDATE Users
 		SET EmailToken = '', EmailVerified = 1
 		WHERE EmailToken = ?`, token.String())
 	return err
 }
 
+// PasswordResetTokenDelete delets the PasswordReset row identified by token
```
jholdstock: deletes*

helpers/auth.go
```diff
@@ -62,18 +69,23 @@ func EmailVerificationTokenExists(dbMap *gorp.DbMap, token models.UserToken) (*m
 	return &user, err
 }
 
+// EmailVerificationComplete Changes a users EmailVerified value to 1.
```
jholdstock: changes*

helpers/auth.go
```diff
@@ -62,18 +69,23 @@ func EmailVerificationTokenExists(dbMap *gorp.DbMap, token models.UserToken) (*m
...
 }
 
+// PasswordResetTokenDelete delets the PasswordReset row identified by token
+// from the DB.
 func PasswordResetTokenDelete(dbMap *gorp.DbMap, token models.UserToken) error {
 	_, err := dbMap.Exec("DELETE FROM PasswordReset WHERE Token = ?", token.String())
 	return err
 }
 
+// PasswordResetTokenExists checks whether the token exits and returns
```
jholdstock: exists*

models/user.go
```diff
@@ -200,20 +211,22 @@ func GetUserCountActive(dbMap *gorp.DbMap) int64 {
 	return userCountActive
 }
 
+// InsertEmailChange inserts a new EmailChange row into the DB.
 func InsertEmailChange(dbMap *gorp.DbMap, emailChange *EmailChange) error {
 	return dbMap.Insert(emailChange)
 }
 
-// InsertLowFeeTicket inserts a user into the DB
+// InsertLowFeeTicket inserts a user into the DB.
```
jholdstock: Doesn't insert users

system/core.go
JoeGruffins: My first thought was that this is needed to satisfy some interface, but I'm not sure what.

system/sqlstore.go
JoeGruffins: At the time I made the sql store, I saw this log quite a bit.  When an operator first updates from whatever to a version using the sql store, depending on the number of users, this log may flood their stdout.  It's not really a problem, it's finding the old type of cookie in the browser, that we can't decode in this fashion, which soon gets replaced. That's why I don't think this log is useful to anyone that's not debugging.  I have not seen the log since then.
Is it better to stay info? It might even be changed to warn once everyone has updated...

JoeGruffins:
@jholdstock 
also, I noticed that there is a method *Controller.Parse and a function helpers.Parse. There only difference is returning an error or logging an error. Seems a little redundant to have both. What do you think?

system/sqlstore.go
jholdstock: Thanks for the explanation. I think that is a decent reason to change the log level to Debug.

jholdstock:
Sounds like a nice thing to fix - removing duplication is almost always a good thing.

jholdstock:
This looks good to merge, but still marked as draft

PR #586, opened 2020-01-21
Parameter to configure database connection encryption

isuldor:
`--dbtls` is a configurable parameter to set the go-sql-driver/mysql option for tls. [1]

1: https://github.com/go-sql-driver/mysql#tls

### DIFF
--- a/backend/stakepoold/config.go
+++ b/backend/stakepoold/config.go
@@ -39,6 +39,7 @@ var (
 	defaultDBName = "stakepool"
 	defaultDBPort = "3306"
 	defaultDBUser = "stakepool"
+	defaultDBTLS  = "false"
 )
 
 // runServiceCommand is only set to a real function on Windows.  It is used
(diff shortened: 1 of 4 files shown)


JoeGruffins:
I don't need to supply a certificate?

isuldor:
I'm glad you brought that up because I'm not sure, and frankly puzzled as to why go's mysql driver would require that with a valid certificate. On the one hand, a simple test script with `http.Get("https://test.stakey.net/")` will happily consume the letsencrypt cert.  But running stakepoold with `--dbtls=true --dbhost=test.stakey.net` results in a `[ERR] DB: Unable to establish connection to db: x509: certificate signed by unknown authority`. This is while the standard mysql client connects with `--ssl-verify-server-cert=on` just fine. It's as if the mysql driver doesn't use the same root store as the http library.

Either way, adding a configurable pem file with [mysql.RegisterTLSConfig](https://github.com/go-sql-driver/mysql#tls) for folks using self-signed certs makes sense. I also omitted it since I'm not exactly sure where and how it should be implemented.

JoeGruffins:
This looks good to me, but I would like to enable custom certificates using the code you linked.  If you don't feel comfortable doing it, I can in another pr.

jholdstock:
This PR only adds support for stakepoold, but dcrstakepool also connects to the db. We should probably add support for both.

dajohi:
@isuldor https://godoc.org/github.com/go-sql-driver/mysql#RegisterTLSConfig

PR #587, opened 2020-01-23
1.5.0 release note

jholdstock:


docs/release-note-1.5.0.md
```diff
@@ -0,0 +1,168 @@
...
+1. Build dcrwallet and dcrd v1.5.0
+1. Announce maintenance and shut down dcrstakepool
+1. Perform an upgrade of each back-end server, one at a time
+   1. Stop stakepoold
+   1. Stop dcrwallet
+   1. Stop dcrd
+   1. Install latest dcrd binary and start
+   1. Install latest dcrwallet binary and start
+   1. Make required changes to stakepoold.conf (detailed [above](#config-changes))
+   1. Install latest stakepoold binary and start
```
JoeGruffins: I feel like we should say something about not starting stakepoold until dcrd and dcrwallet have synced with the network.  I think we had a problem here last time everyone updated. Or is there anything we could do in the code? should make an issue?

jholdstock:
Thanks Joe. Implemented both suggestions.

I actually thought stakepoold could handle the scenario where dcrd/dcrwallet are not fully up to date. This may be worth investigating further, because it really should be able to handle it.

JoeGruffins:
I think the scenario was thus, the operator updated their vsp, they started everything at once. Right after the pool coming online, a new user links their account to a wallet and buys some tickets, the wallet is currently syncing, and the redeem script is not imported until the initial sync is over. However, the bought tickets are in that initial sync, so the wallet passes them over. I will try to reproduce this, I'll make an issue if I can.  The conversation about this is in the pos room on Tue, Nov 26 2019

JoeGruffins:
So far unable to reproduce... iirc the solution was to rescan their wallets.  So they had the redeem script, but for some reason the wallets, all of them, had missed the ticket purchase.  It was like 10 tickets purchased by the same new user.

jholdstock:
Well it's included in the steps now anyway. Better to be safe than sorry. Hopefully it prevents somebody else running into issues.

PR #589, opened 2020-01-29
Remove unnecessary item from sample config.

jholdstock:
Also increase the maximum ticket price because the price is always above 100 these days

### DIFF
--- a/views/address.html
+++ b/views/address.html
@@ -117,7 +117,6 @@
 									poolfees={{ .PoolFees }} <br>
 									[Ticket Buyer Options] <br>
 									ticketbuyer.votingaddress={{ .User.MultiSigAddress }} <br>
-									ticketbuyer.maxpriceabsolute=200
 								</p>
 							</div>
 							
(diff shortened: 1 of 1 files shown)


isuldor:
iirc `--ticketbuyer.maxpriceabsolute` was deprecated after the stakediff fork when ticketbuyer was overhauled.  dcrwallet will not start when that option is set due to it being invalid.

views/address.html
```diff
@@ -115,11 +115,9 @@ <h1 class="modal-title" id="connectSubAddressTitle"><img src="/assets/images/inf
 									enableticketbuyer=true <br>
 									pooladdress={{ .User.UserFeeAddr }} <br>
 									poolfees={{ .PoolFees }} <br>
-									;; DEPRECATED -- use ticketbuyer.votingaddress instead <br>
-									;; ticketaddress={{ .User.MultiSigAddress }} <br>
 									[Ticket Buyer Options] <br>
 									ticketbuyer.votingaddress={{ .User.MultiSigAddress }} <br>
-									ticketbuyer.maxpriceabsolute=100
+									ticketbuyer.maxpriceabsolute=200
```
isuldor: iirc --ticketbuyer.maxpriceabsolute was deprecated after the stakediff fork when ticketbuyer was overhauled. dcrwallet will not start when that option is set due to it being invalid.  This should probably be removed.

jholdstock: Good catch. I've updated the PR

PR #595, opened 2020-02-19
build: drop go 1.12, add go 1.14

dajohi:


jholdstock:
lgtm after README is updated

PR #601, opened 2020-03-02
controllers/main: Add agendas test

JoeGruffins:
To facilitate testing, change agendas to accept an optional function
that takes the place of dcrDataAgendas.

Return a pointer to agendas rather than making a copy.

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -185,24 +185,29 @@ func (controller *MainController) getNetworkName() string {
 
 // agendas returns agendas and their statuses. Fetches agenda status from
 // dcrdata.org if past agenda.Timer limit from previous fetch. Caches agenda
-// data for agendasCacheLife. This method is safe for concurrent use.
-func (controller *MainController) agendas() []agenda {
+// data for agendasCacheLife. Takes an infos function that can be used to
+// fetch agenda info. Defaults to dcrDataAgendas. This method is safe for
+// concurrent use.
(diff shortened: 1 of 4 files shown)


controllers/main.go
```diff
@@ -1961,7 +1966,7 @@ func (controller *MainController) Voting(c web.C, r *http.Request) (string, int)
 		c.Env["Agenda"+strk+"Selected"] = v
 	}
 	c.Env["Admin"], _ = controller.isAdmin(c, r)
-	c.Env["Agendas"] = controller.agendas()
+	c.Env["Agendas"] = controller.agendas(nil)
```
jholdstock: I don't really like this change, passing `nil` into this function is non-intuitive. On top of that, passing a function into a function is unnecessary complex for what we are trying to achieve here. The small gain in test coverage isn't worth the extra complexity imo.

Compromising the quality of production code in order to support testing code is usually a code smell. Perhaps this code needs a bit of refactoring before some tests are added (if we really even need tests in this area).

JoeGruffins: Ok, will look into refactoring. Any ideas to make it cleaner?

JoeGruffins: Not refactored but product code is unchanged now.

JoeGruffins:
@dajohi poke

JoeGruffins:
@dajohi rebased

PR #602, opened 2020-03-03
controllers/main: Add APIAddress test

JoeGruffins:


jholdstock:
Will review this after #598 is merged. Feel free to give me a poke if 598 goes in and I forget to revisit this

JoeGruffins:
@jholdstock **POKE**

controllers/main_test.go
```diff
@@ -445,3 +459,254 @@ func TestNewMainController(t *testing.T) {
...
+		c.Env["APIUserID"] = test.userID
+		// nil APIUserID indicats an authentication error occured.
+		// Simulating that on userid -1.
+		if test.userID == -1 {
+			c.Env["APIUserID"] = nil
+		}
+		c.Env["DbMap"] = dbMap
+		// Put userPubKeyAddr as form value.
+		r, _ := http.NewRequest("GET", "?UserPubKeyAddr="+test.userPubKeyAddr, nil)
+		_, errCode, _, _ := mc.APIAddress(c, r)
```
jholdstock: I wonder if we should also be checking the other values returned by `APIAddress()`.

This actually makes me wonder, do we really need to return both a string response and an error? The error alone is probably sufficient, having both seems redundant.

PR #606, opened 2020-03-08
stakepoold: Deprecate wallet RPC ListScripts

JoeGruffins:
Use UserVotingConfig keys and RPC command ValidateAddress instead.

### DIFF
--- a/backend/stakepoold/stakepool/stakepool.go
+++ b/backend/stakepoold/stakepool/stakepool.go
@@ -361,11 +361,6 @@ func (spd *Stakepoold) AddMissingTicket(ticketHash []byte) error {
 // and returns a list of redeem scripts.
 func (spd *Stakepoold) ListScripts() ([][]byte, error) {
 	const op = "ListScripts"
-	type result struct {
-		addr, hex string
-		err       error
-	}
-	results := make(chan *result)
 	spd.Lock()
 	addrs := make([]dcrutil.Address, 0, len(spd.UserVotingConfig))
 	// UserVotingConfig should have multi sig addresses for all users.
@@ -380,6 +375,11 @@ func (spd *Stakepoold) ListScripts() ([][]byte, error) {
 		addrs = append(addrs, decodedAddress)
 	}
 	spd.Unlock()
+	type result struct {
+		addr, hex string
+		err       error
+	}
+	results := make(chan *result)
 	var wg sync.WaitGroup
 	wg.Add(len(addrs))
 	// listScripts validates a slice of addresses using RPC calls.
--- a/backend/stakepoold/stakepool/stakepool.go
+++ b/backend/stakepoold/stakepool/stakepool.go
@@ -361,30 +361,78 @@ func (spd *Stakepoold) AddMissingTicket(ticketHash []byte) error {
 // and returns a list of redeem scripts.
 func (spd *Stakepoold) ListScripts() ([][]byte, error) {
 	const op = "ListScripts"
+	type result struct {
+		addr, hex string
+		err       error
+	}
+	results := make(chan *result)
 	spd.Lock()
-	defer spd.Unlock()
-	scripts := make([][]byte, 0, len(spd.UserVotingConfig))
+	addrs := make([]dcrutil.Address, 0, len(spd.UserVotingConfig))
+	// UserVotingConfig should have multi sig addresses for all users.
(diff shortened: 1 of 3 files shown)


backend/stakepoold/stakepool/stakepool.go
```diff
@@ -356,15 +356,82 @@ func (spd *Stakepoold) AddMissingTicket(ticketHash []byte) error {
...
+		for _, addr := range addrs {
+			defer wg.Done()
+			validated, err := spd.WalletConnection.RPCClient().ValidateAddress(addr)
+			if err != nil {
+				err = fmt.Errorf("failed to validate address %s", addr.String())
+				log.Error("%s: %v", op, err)
+				results <- &result{err: err}
+				continue
+			}
+			results <- &result{addr: addr.String(), hex: validated.Hex}
```
jholdstock: Why is this using `validated.Hex` rather than `validated.Script`?

JoeGruffins: iirc the type of script is there, not the script, for example, "multisig"

JoeGruffins: https://github.com/decred/dcrwallet/blob/270804fbd664b5cb124400c3ea10dcdb21bb51de/internal/rpc/jsonrpc/methods.go#L3620

jrick:
functionally this looks fine and it works in a pinch when using the most recent wallet but i don't believe this is the best way to to about this.  ideally it would be comparing addresses across wallets, and only looking up the scripts for addresses that some wallet was missing.  getaddressesbyaccount appears broken on the imported account in dcrwallet right now however...

doesn't stakepoold already have a complete list of all addresses it should be watching for ticket purchases? could that be used instead?

JoeGruffins:
@jrick We have to check to make sure that each wallet has all user redeem scripts. Are you saying that there's a better way to check than validateaddress? If so we will use that.

jrick:
there should be, but getaddressesbyaccount seems broken for the imported account currently.  if that worked, you could easily check for all p2sh addresses for each script.

in lieu of that, you can perform validateaddress calls checking to see if each result for the p2sh addresses has `ismine: true`.  if one is not, these can be recorded per-wallet, fetching the from other sources (e.g. other wallets which had them perhaps), and imported.

JoeGruffins:
Created an issue #613 

Also, I just realized that stakepoold works without a connection to the database. I think that needs to be changed.

JoeGruffins:
@jrick 
> in lieu of that, you can perform validateaddress calls checking to see if each result for the p2sh addresses
Are they multisig or p2sh addresses? They are two different things, correct? do multisig need to be followed by a p2sh script to be spent?

jrick:
Currently, the ticket outputs are p2sh. the p2sh redeem script is a 1-of-2 multisig.

JoeGruffins:
@jholdstock @dajohi 
Unless I'm mistaken, master is currently broken without these changes because #609 has been merged?

itswisdomagain:
> there should be, but getaddressesbyaccount seems broken for the imported account currently. if that worked, you could easily check for all p2sh addresses for each script.
> 
> in lieu of that, you can perform validateaddress calls checking to see if each result for the p2sh addresses has `"ismine": true`. if one is not, these can be recorded per-wallet, fetching the script from other sources (e.g. other wallets which had them perhaps), and imported.

dcrstakepool has the list of p2sh addresses it watches, together with the redeem script for each. Calling `validateaddress` for each address per stakepoold instance/wallet would be a lot of rpc calls being made when dcrstakepool starts up, all just to determine which wallet is missing which script, if any.

I don't know if it was discussed before, the possibility of passing multiple addresses to the `validateaddress` rpc method. If that's doable, the number of rpc calls to check for missing scripts would reduce to perhaps 1 per stakepoold instance/wallet.

Might be a better idea to fix `getaddressesbyaccount` for imported accounts and use that instead, to reduce the number of rpc calls to 1 per wallet (excluding calls to import missing scripts).

jrick:
I forgot to update here, but getaddressesbyaccount on the imported account is fixed in dcrwallet master.

JoeGruffins:
will try to solve #613 in another pr, using getaddressesbyaccount.

JoeGruffins:
#619 which uses getaddressesbyaccount should be considered instead of this pr

PR #609, opened 2020-03-09
stakepool: bump minimum dcrwallet api requirement

dajohi:


jrick:
fyi, version 7 will also remove the listscripts method (decred/dcrwallet#1688).  #606 will be needed after that is removed.

JoeGruffins:
Tested with https://github.com/decred/dcrwallet/pull/1688 and #606 and no problems for me.

PR #610, opened 2020-03-09
multi: use std errors package

### LINKED ISSUE #518: Add errors package

dajohi:
Closes #518

### DIFF
--- a/backend/stakepoold/config.go
+++ b/backend/stakepoold/config.go
@@ -1,11 +1,12 @@
 // Copyright (c) 2013-2014 The btcsuite developers
-// Copyright (c) 2015-2019 The Decred developers
+// Copyright (c) 2015-2020 The Decred developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
 package main
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"os"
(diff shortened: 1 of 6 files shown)


config.go
```diff
@@ -362,7 +364,8 @@ func loadConfig() (*config, []string, error) {
 	if !(preCfg.SimNet) || preCfg.ConfigFile != defaultConfigFile {
 		err := flags.NewIniParser(parser).ParseFile(preCfg.ConfigFile)
 		if err != nil {
-			if _, ok := err.(*os.PathError); !ok {
+			var e *os.PathError
+			if !errors.As(err, &e) {
```
JoeGruffins: Can use Is since we don't care about the unwrapped value.

backend/stakepoold/config.go
```diff
@@ -340,7 +342,8 @@ func loadConfig() (*config, []string, error) {
 	if !(preCfg.SimNet) || cfg.ConfigFile != defaultConfigFile {
 		err := flags.NewIniParser(parser).ParseFile(cfg.ConfigFile)
 		if err != nil {
-			if _, ok := err.(*os.PathError); !ok {
+			var e *os.PathError
+			if !errors.As(err, &e) {
```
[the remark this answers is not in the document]
dajohi: We care about the type, so we use As().  Is() matches values.

config.go
dajohi: We care about the type, so we use As().  Is() matches values.

Issue #495, opened 2019-08-21
stakepoold access to MySQL db data without needing to provide MySQL connection params on all stakepoold instances


itswisdomagain:
Currently, stakepoold accesses the MySQL db running on the dcrstakepool web server to read user voting preferences and added low fee tickets. This db read op is performed on 2 occasions: when stakepoold starts up and every couple minutes (if stakepoold gRPC server is disabled).

Assuming that stakepoold gRPC server is enabled (shouldn't it always?), we can add a stakepoold gRPC method to supply the above data to all connected stakepoold backends when dcrstakepool starts.

This has the advantage of not needlessly exposing the MySQL db to external IPs (considering that the db would typically sit on the web server, can just keep access to the db local to the running dcrstakepool instance).

It also simplifies the process for setting up a vsp, as vsp admins would have 4 fewer stakepoold config fields to worry about when setting up their backends.

PS: The implication is that the stakepoold `norpclisten` config option will have to be deprecated. `dcrstakepool` will currently not even start if the stakepoold backends do not have their rpc servers enabled.

JoeGruffins:
Might as well move to boltdb then.

jholdstock:
As mentioned on matrix, supplying config through RPC will always get a firm No from me. Servers should be able to run in isolation and should not depend on config values being passed in via RPC from another server.

itswisdomagain:
The data I'm proposing to supply through RPC are not config values, they're actually currently being passed via RPC when the data changes in the database, but not when dcrstakepool starts up.

https://github.com/decred/dcrstakepool/blob/master/controllers/main.go#L520
https://github.com/decred/dcrstakepool/blob/master/controllers/main.go#L537

JoeGruffins:
I agree with @itswisdomagain This is a good idea. We don't have to expose the db, and it simplifies things.  It paves the way for maybe a db change in the future.

itswisdomagain:
@jholdstock what's your thought on this?

itswisdomagain:
Based on recent conversation on matrix, particularly @jholdstock's message [here](https://matrix.to/#/!HEeJkbPRpAqgAwhXWO:decred.org/$156950220439939nBeEZ:decred.org?via=decred.org&via=matrix.org&via=pintobyte.com), it makes sense to continue to allow stakepoold direct access to MySQL, to enable stakepoold run as independently from dcrstakepool as possible.

chappjc:
[...]

PR #618, opened 2020-03-23
multi: Move script syncing to stakepoold

### LINKED ISSUE #613: stakepoold: Make startup redeem script checking more efficient
More discussion here: #606 

Currently scripts are gathered from all wallets and compared against each other. It should be possible to get all addresses and scripts from the database and have stakepoold compare addresses fetched from dcrwallet to the database results and import the scripts they need independently. I do not think stakepooldclient needs to do any work at all. Only stop on and report errors. 

stakepoold will work without having a connection to a database, which must be changed in order for this to be effective.

The code in question is here:

https://github.com/decred/dcrstakepool/blob/3394b5069f4b459435e5c9fcfb56ecd1929a767c/stakepooldclient/stakepooldclient.go#L348-L435

and stakepoold's ListScripts command.

JoeGruffins:
closes #613  

Remove grpc methods ImportMissingScripts and ListScripts. Stakepoold now
require a connection to a mysql database in order obtain user
preferences which include a list of redeem scripts and heights
registered. Upon start-up, all scripts found in the database but not in
dcrwallet are imported. Remove user config save file as all data will
come from the database.

This is an alternative to ~#606~ #619


chappjc:
I always thought we wanted to get rid of the stakepoold-mysql connection.  stakepoold already requires a mysql connection, no?  I'm confused by the comment in https://github.com/decred/dcrstakepool/pull/606#issuecomment-597964050 because the last release of stakepoold definitely connects to mysql, and I never liked that.

chappjc:
Is a stakepoold-mysql connection only used presently when `--norpclisten` is enabled?

chappjc:
I'm strongly in favor of https://github.com/decred/dcrstakepool/pull/606 if it works too.  What are the pros and cons of each solution (this PR vs. https://github.com/decred/dcrstakepool/pull/606).

chappjc:
See also https://github.com/decred/dcrstakepool/issues/495, which I think makes a very strong case, and https://github.com/decred/dcrstakepool/issues/134 for historical context.

JoeGruffins:
I left my thoughts about connecting mysql and stakepoold in #495 

You can see in this code that cutting out the middle man for this one instance has also cut the needed code in half.

However, if this indeed causes a security issue, I will work on another implementation using addressesbyaccount. Security is most important.

itswisdomagain:
> However, if this indeed causes a security issue, I will work on another implementation using addressesbyaccount. Security is most important.

I made an implementation using `getaddressesbyaccount` a while ago. I think I'll just create a PR for it and throw it into the mix for consideration.

PR #619, opened 2020-03-30
stakepoold syncScripts: replace ListScripts with ListImportedAddresses

itswisdomagain:
stakepoold relies on dcrwallet's listscripts JSON-RPC method to determine multisig scripts that are missing from a stakepoold's wallet. The `listscripts` rpc method was removed in https://github.com/decred/dcrwallet/pull/1688 which now causes stakepoold to fail on start.

This PR replaces stakepoold's usage of the `listscripts` rpc method with `getaddressesbyaccount` for the imported account. Both rpc methods serve the same purpose on stakepoold - determine multisig scripts that are missing from a stakepoold's wallet.

The `listscripts` method returns a list of scripts that are found in a stakepoold's wallet, and compares this with the scripts in the dcrstakepool database to determine which scripts are yet to be imported into the stakepoold instance.

The `getaddressesbyaccount "imported"` method returns a list of addresses found in the `imported` account of a stakepoold's wallet, which is then compared with the list of multisig addresses in the dcrstakepool database to determine which scripts are yet to be imported into the stakepoold instance.


backend/stakepoold/rpc/stakepoolrpc/api.pb.go
```diff
@@ -1782,112 +1782,112 @@ func init() {
...
 
 // Reference imports to suppress errors if they are not otherwise used.
 var _ context.Context
-var _ grpc.ClientConnInterface
+var _ grpc.ClientConn
 
 // This is a compile-time assertion to ensure that this generated file
 // is compatible with the grpc package it is being compiled against.
-const _ = grpc.SupportPackageIsVersion6
+const _ = grpc.SupportPackageIsVersion4
```
JoeGruffins: could you update your protoc-gen-go to 1.3.4 or later

backend/stakepoold/stakepool/stakepool.go
```diff
@@ -357,15 +357,21 @@ func (spd *Stakepoold) AddMissingTicket(ticketHash []byte) error {
...
 		return nil, err
 	}
 
-	return scripts, nil
+	addrStrings := make([]string, len(addresses))
+	for i, address := range addresses {
+		addrStrings[i] = address.String()
+	}
+
+	return addrStrings, nil
```
JoeGruffins: I think addresses is already a []string and you can just return that.

JoeGruffins:
could you also put closes #613 in the title

Issue #574, opened 2019-12-11
Design and API proposal for accountless operation


jrick:
This issue describes at a high level an alternate design to #568 to solve the privacy issues associated with VSP accounts, while simultaneously removing the VSP fees from tickets themselves, making it possible for both solo and VSP stakers to purchase mixed tickets in the same anonymity set.

As a reminder, BIP0044 with Decred's coin types uses the following HD paths:

`m/44'/{20,42}'/<account>'/<branch>/<child>`

For this design, a new purpose key should be used, since these keys are used differently from normal wallet accounts, and handling should be logically and hierarchically separated.  A new purpose key of 565350 is chosen (565350 is the UTF-8 encoding of "VSP", as hex).

For each bought ticket, a hardened key is derived.  Non-hardened children are derived off this ticket key.  The 0th hardened child is used as a key for voting rights in the ticket purchase, and more child keys may be derived for future uses as we need them.

Tickets will be purchased with voting rights assigned to a P2PKH address derived from the key:

`m/565350'/<ticket index>'/0'`

and a commitment address from one of the wallet's BIP0044 accounts.

There will be only one pair of commitment/change outputs for the ticket.  Currently, VSP-bought tickets include two pairs of these outputs: one for the buyer's reward and a second to pay the VSP a fee for voting on behalf of the user.  Removing these two separate outputs, and switching voting rights from a (multisig) P2SH to P2PKH has the desired effect of making these tickets have identical form as those bought by solo stakers, and VSP users will be able to participate in the same anonymity set as solo buyers when using CoinShuffle++ to create CoinJoin mixed ticket split transactions, from which tickets are bought.

Between the ticket being mined and becoming live, the user (or their client) must inform the VSP of the ticket to vote on, the private key necessary to vote, voting preferences, and pay the VSP fee.  The VSP fee can be obtained via an API and should be checked prior to buying the ticket, so that an appropriate UTXO can be reserved by the wallet to pay the fee, without that output being double spent by the created split tx.

(These mock APIs loosely resemble JSON-RPC, but HTTP REST is probably more appropriate, and is already used by the existing API.)

```
client request:
{
	"request": "getfee"
}
server response:
{
	"fee": number
}
```

[...]

PR #625, opened 2020-05-07
[WIP] APIv3

dajohi:
Work towards #574


controllers/main.go
```diff
@@ -2119,6 +2160,220 @@ func (controller *MainController) IsValidVoteBits(userVoteBits uint16) bool {
...
+	return &poolapi.GetPubKeyResponse{
+		Timestamp: time.Now().Unix(),
+		PubKey:    controller.Cfg.PubKey,
+	}
+}
+
+func (controller *MainController) APIv3GetFee(c web.C, r *http.Request) *poolapi.GetFeeResponse {
+	return &poolapi.GetFeeResponse{
+		Timestamp: time.Now().Unix(),
+		Fee:       controller.Cfg.PoolFees,
```
jholdstock, line 2173: I believe this should be returning an absolute value, the exact fee for the ticket specified in the request.

controllers/main.go
```diff
@@ -2119,6 +2160,220 @@ func (controller *MainController) IsValidVoteBits(userVoteBits uint16) bool {
...
+	}
+}
+
+func (controller *MainController) APIv3GetFeeAddress(c web.C, r *http.Request) (*poolapi.GetFeeAddressResponse, error) {
+	// HTTP GET Params required
+	// ticketHash - hash of ticket
+	// signature - signmessage signature using the ticket commitment address
+	//	     - message = "vsp v3 getfeeaddress ticketHash"
```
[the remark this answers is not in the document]
itswisdomagain: Something that was considered in the design of #514 is avoiding replay attacks.
- If ssl is not used, an attacker would be able to capture the signature and use it maliciously.
- Even if ssl is used, an attacker could still resend the request as is, without needing to decrypt it, and would be able to get the server to use an outdated voting pref from an older request in the event that a user has recently changed their prefs.

controllers/main.go
```diff
@@ -2119,6 +2160,220 @@ func (controller *MainController) IsValidVoteBits(userVoteBits uint16) bool {
...
+	votingKey := r.FormValue("votingKey")
+	votingWIF, err := dcrutil.DecodeWIF(votingKey, controller.Cfg.NetParams.PrivateKeyID)
+	if err != nil {
+		return nil, err
+	}
+
+	feeTxStr := r.FormValue("feeTx")
+	feeTxBytes, err := hex.DecodeString(feeTxStr)
+	if err != nil {
+		return nil, errors.New("invalid transaction")
```
itswisdomagain, line 2266:
```suggestion
		return nil, errors.New("invalid fee transaction")
```

controllers/main.go
```diff
@@ -2119,6 +2160,220 @@ func (controller *MainController) IsValidVoteBits(userVoteBits uint16) bool {
...
+					feeAddr = validFeeAddr
+					feeAmount = dcrutil.Amount(txOut.Value)
+					break findAddress
+				}
+			}
+		}
+	}
+	if feeAddr == "" {
+		log.Infof("feeTx did not invalid any payments")
+		return nil, errors.New("feeTx did not include any payments")
```
itswisdomagain, line 2316:
```suggestion
		log.Infof("feeTx does not include a fee payment output")
		return nil, errors.New("feeTx does not include a fee payment output")
```

controllers/main.go
```diff
@@ -2119,6 +2160,220 @@ func (controller *MainController) IsValidVoteBits(userVoteBits uint16) bool {
...
+	}
+
+	feeEntry, err := models.GetFeesByFeeAddress(dbMap, feeAddr)
+	if err != nil {
+		log.Infof("GetFeeByAddress: %v", err)
+		return nil, errors.New("database error")
+	}
+	voteAddr, err := dcrutil.DecodeAddress(feeEntry.Address, controller.Cfg.NetParams)
```
itswisdomagain, line 2324:
```suggestion
	commitmentAddr, err := dcrutil.DecodeAddress(feeEntry.Address, controller.Cfg.NetParams)
```

controllers/main.go
```diff
@@ -2119,6 +2160,220 @@ func (controller *MainController) IsValidVoteBits(userVoteBits uint16) bool {
...
+	}
+
+	minFee := txrules.StakePoolTicketFee(sDiff, relayFee, int32(feeEntry.BlockHeight), controller.Cfg.PoolFees, controller.Cfg.NetParams)
+	if feeAmount < minFee {
+		log.Infof("too cheap: %v %v", feeAmount, minFee)
+		return nil, fmt.Errorf("dont get cheap on me, dodgson (sent:%v required:%v)", feeAmount, minFee)
+	}
+
+	// Get vote tx to give to wallet
+	ticketHash, err := chainhash.NewHashFromStr(feeEntry.TicketHash)
```
itswisdomagain, line 2352: comment

controllers/main.go
```diff
@@ -2119,6 +2160,220 @@ func (controller *MainController) IsValidVoteBits(userVoteBits uint16) bool {
...
+	if err != nil {
+		log.Errorf("PayFee: %v", err)
+		return nil, errors.New("RPC server error")
+	}
+
+	err = models.InsertFeeAddressVotingKey(dbMap, voteAddr.Address(), votingWIF.String(), voteBits)
+	if err != nil {
+		log.Errorf("PayFee: InsertVotingKey failed: %v", err)
+		return nil, errors.New("internal error")
+	}
```
itswisdomagain, line 2369: To consider: what can a user do to get refunded if this happens? Since anyone could have broadcasted the fee tx, how does the user know for certain that the tx was broadcasted by the VSP to demand refund?

controllers/main.go
```diff
@@ -2119,6 +2160,220 @@ func (controller *MainController) IsValidVoteBits(userVoteBits uint16) bool {
...
+	resp, err := controller.Cfg.StakepooldServers.PayFee(r.Context(), ticketHash, votingWIF, dcrutil.NewTx(feeTx))
+	if err != nil {
+		log.Errorf("PayFee: %v", err)
+		return nil, errors.New("RPC server error")
+	}
+
+	err = models.InsertFeeAddressVotingKey(dbMap, voteAddr.Address(), votingWIF.String(), voteBits)
+	if err != nil {
+		log.Errorf("PayFee: InsertVotingKey failed: %v", err)
+		return nil, errors.New("internal error")
```
itswisdomagain, line 2368: I've seen return `internal error`, `internal database error` and `database error`. Could harmonize all into a reusable `var dbErr = errors.New("internal database error")` or a more generic `var internalServerErr = errors.New("internal error")`.

models/tickets.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"github.com/go-gorp/gorp"
+	_ "github.com/go-sql-driver/mysql"
+)
+
+type Fees struct {
+	TicketHash          string
+	CommitmentSignature string
+	FeeAddress          string
+	Address             string
```
itswisdomagain, line 16: `CommitmentAddress`. Prolly immediately before or after `CommitmentSignature`.

models/tickets.go
```diff
@@ -0,0 +1,74 @@
...
+	return addrs, err
+}
+
+// InsertFeeAddress inserts a fee address into the DB.
+func InsertFeeAddress(dbMap *gorp.DbMap, fee *Fees) error {
+	return dbMap.Insert(fee)
+}
+
+// InsertFeeAddressVotingKey updates the fees table with a voting key for the given address.
+func InsertFeeAddressVotingKey(dbMap *gorp.DbMap, address, votingKey string, voteBits uint16) error {
```
itswisdomagain, line 68:
```suggestion
func SetFeeAddressVotingKey(dbMap *gorp.DbMap, address, votingKey string, voteBits uint16) error {
```

models/tickets.go
```diff
@@ -0,0 +1,74 @@
...
+	_, err := dbMap.Exec("UPDATE Fees SET VotingKey = ?, VoteBits = ? WHERE Address = ?", votingKey, voteBits, address)
+	if err != nil {
+		log.Warnf("InsertFeeAddressPrivKey: %v", err)
```
itswisdomagain, line 71: Needs update.

stakepooldclient/stakepooldclient.go
```diff
@@ -731,3 +734,41 @@ func (s *stakepooldManager) CrossCheckColdWalletExtPubs(ctx context.Context, dcr
...
+		client := pb.NewStakepooldServiceClient(conn)
+		responses[i], err = client.GetFeeAddress(ctx, &pb.GetFeeAddressRequest{
+			Hash:      txHash.String(),
+			Signature: signature,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("GetFeeAddress RPC failed on stakepoold instance %s: %v", conn.Target(), err)
+		}
+	}
+	return responses[0], nil
```
itswisdomagain, line 751: 2 questions here:
- This method errors if any of the voting wallets do not return an address but the addresses returned from all backends are not compared for consistency. If it doesn't matter whether they all return the same address, I suppose it should likewise not matter if 1 or more backends do not return an address.
- If a previous call to this method ends up deriving a new address for some wallets and not others, a new call to the method would produce different addresses. I expect the wallet that failed previously would be an index behind and would not be tracking usage for the newer address. I think that just means the balance may not be accurate on that wallet, so no biggie perhaps. If no biggie, could just ignore `GetFeeAddress RPC failure` on any failing backend.

I still think using a cold wallet's xpub is better anyways.

stakepooldclient/stakepooldclient.go
```diff
@@ -731,3 +734,41 @@ func (s *stakepooldManager) CrossCheckColdWalletExtPubs(ctx context.Context, dcr
...
+		responses[i], err = client.PayFee(ctx, &pb.PayFeeRequest{
+			TicketHash: txHash.String(),
+			VotingWIF:  votingWIF.String(),
+			FeeTx:      feeTxBuffer.Bytes(),
+		})
+		if err != nil {
+			return nil, fmt.Errorf("GetFeeAddress RPC failed on stakepoold instance %s: %v", conn.Target(), err)
+		}
+	}
+	return responses[0], nil
```
itswisdomagain, line 773: 2 concerns here
- Calling `PayFee` on multiple backends would cause multiple fee tx broadcast attempts. The broadcast could fail on any of those backends, no?
- If `PayFee` errors on one backend, the fee tx might have been broadcasted by another backend. Should probably ignore errors if the `PayFee` call to the first backend succeeds.

controllers/main.go
itswisdomagain: should **not** be returning an absolute value, you mean?

jholdstock: Yep, you're right

backend/stakepoold/stakepool/stakepool.go
```diff
@@ -468,6 +468,124 @@ func (spd *Stakepoold) ValidateAddress(ctx context.Context, address string) (*wa
...
+		return nil, nil, 0, 0, err
+	}
+	blockHeader, err := spd.NodeConnection.GetBlockHeader(ctx, blockHash)
+	if err != nil {
+		return nil, nil, 0, 0, err
+	}
+	sDiff := blockHeader.SBits
+
+	// get fee address
+	feeAddress, err := spd.WalletConnection.RPCClient().GetNewAddress(ctx, "fees")
```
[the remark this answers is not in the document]
JoeGruffins: Maybe it will be an imported account? ie, `importxpub "name" "xpub"` name being "fee"

config.go
```diff
@@ -593,5 +597,26 @@ func loadConfig() (*config, []string, error) {
...
+			return nil, nil, fmt.Errorf("failed to generate "+
+				"signing key: %v", err)
+		}
+		err = ioutil.WriteFile(seedPath, cfg.signKey.Seed(), 0400)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to save "+
+				"signing seed: %v", err)
+		}
+	} else {
+		cfg.signKey = ed25519.NewKeyFromSeed(seed)
```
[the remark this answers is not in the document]
JoeGruffins: I agree that this should probably be in the db, if there will be a db, and backed up. A random file is easy to delete.

backend/stakepoold/stakepool/stakepool.go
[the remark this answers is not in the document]
dajohi: yes, there will be an importxpub

stakepooldclient/stakepooldclient.go
dajohi: sql going away

itswisdomagain: 🤔 reply meant for another comment perhaps?

xaur:
What is the status of this PR now that [vspd](https://github.com/decred/vspd) is announced?

jholdstock:
https://github.com/decred/vspd

PR #627, opened 2020-05-08
models: change GetDbMap to return an error

dajohi:
system: change Init to a func that returns an Application


jholdstock:
Yeah I would tend to agree on this. The prefixes added to the errors provide an additional bit of info, and also something which can be grepped in our own code base. Rest of the changes look fine

PR #636, opened 2021-05-17
Replace addticket with addtransaction

jrick:
This allows stakepoold to function when deploying dcrwallet v1.6.x.


jholdstock (approved):
Would be good to update the error logging as well, but as this is just dcrstakepool, I really don't care :)

